PR #11, opened 2024-07-31
Add Input Field + Label Components 

milesha:
## Description:
- Added Label Component supporting three sizes
- Added Input Field Component supporting three sizes and disabling
     - Storybook contains an example of using input with a button (disabling the button if the field is empty)
     - Storybook contains an example of using input with a label 

## Screenshot (for UI changes):
Screenshot 2024-07-31 at 21 46 59
Screenshot 2024-07-31 at 21 47 18

## Checklist

- [x] Define assignee and add/update labels
- [x] Check browser console log
- [x] Add Storybook files


webapp/src/app/ui/label/label.component.ts
```diff
@@ -0,0 +1,40 @@
...
+
+@Component({
+  selector: 'app-label',
+  standalone: true,
+  templateUrl: './label.component.html',
+})
+export class AppLabelComponent {
+  class = input<ClassValue>('');
+  for = input<string>('');
+  size = input<LabelVariants['size']>('default');
```
FelixTJDietrich, line 35: I think we don't need the size attribute 🤔 Did you do it in order to have meaningful args? If the helper does not support it you can fix it. I think I would still leave it like this with `cva`, `args`, and `argTypes` in case we want to add variants to the label in the future.

FelixTJDietrich (commented):
Code looks good to me :)

PR #12, opened 2024-07-31
Use Springdoc's OpenAPI Specs for Client

FelixTJDietrich:
## Description

- Remove Pokemon API 
- Generate OpenAPI spec through an springdoc plugin
- Add root level npm package for monorepo scripts
- API generation scripts
  - Add script to clean up client API
  - Add script to generate the `openapi.yaml` for the application server
  - Add script to generate the webapp API client based on the application server specs
  - Add script to execute all commands in order
- Update example using our "hello world" endpoints


FelixTJDietrich:
It would be nice to have a GitHub workflow that automatically generates the OpenAPI specs and client and commits it if it changed.

PR #15, opened 2024-07-31
Automatically update OpenAPI specs and client

FelixTJDietrich:


FelixTJDietrich:
Will remove the automatic commit stuff, the check will just fail if the api client has not been updated.

PR #35, opened 2024-08-06
Fix: CVA Tailwind Autocomplete

GODrums:
#### Summary
This adds the auto-complete feature for Tailwind classes (VSCode's Tailwind IntelliSense extension) to our custom CVA implementation for Angular components.

image


FelixTJDietrich (approved):
Worked after restarting VSCode. 
We might need further fixing for other cases. I checked `computedClass` and it also did not give any IntelliSense when adding classes there.

PR #42, opened 2024-08-08
Update Avatar and improve fallback

FelixTJDietrich:
#### Motivation

Avatar component was fine but I wanted to try out more complex Angular features.

I was thinking of using directives but I still don't like them. 

#### Key Changes

- Split Avatar into multiple components, `app-avatar`, `app-avatar-image`, and `app-avatar-fallback`
- Fallback can be anything now
- Remove CVA from Avatar since it is not needed, we can also do this for future components that don't need variants
- Also inline the html templates, we can do this for small components

#### General

#### Client (if applicable)
- [ ] User experience and accessibility have been tested
- [x] Added Storybook stories for new components

*Attach screenshots here.*


webapp/src/app/ui/avatar/avatar.component.ts
```diff
@@ -1,51 +1,31 @@
...
 })
-export class AppAvatarComponent {
-  class = input<ClassValue>('');
-  size = input<AvatarVariants['size']>('default');
-
-  src = input.required<string>();
-  alt = input<string>('');
-  imageClass = input<string>('');
-  fallback = input<string>('https://placehold.co/56');
+export class AvatarComponent {
```
GODrums, line 22: Nit: for consistency reasons this should probably use the `App`-prefix like our other component classes, but if I learned Angular correctly only the component-selector really matters anyways.

webapp/src/app/ui/avatar/avatar.stories.ts
```diff
@@ -1,95 +1,34 @@
...
-
   render: (args) => ({
     props: args,
-    template: `<app-avatar ${argsToTemplate(args)}></app-avatar>`
+    template: `
+      <app-avatar>
+        <app-avatar-image src="https://github.com/shadcn.png" alt="@shadcn" />
+        <app-avatar-fallback>CN</app-avatar-fallback>
+      </app-avatar>
+    `
```
GODrums, line 32: Just for testing purposes it would be great to be able to test different avatars or the fallback component with the interactive controls in Storybook (being able to change the src attribute would suffice I guess).

image

webapp/src/app/ui/avatar/avatar-config.ts
```diff
@@ -0,0 +1,24 @@
+import { inject, InjectionToken, Provider } from '@angular/core';
+
+export interface AvatarConfig {
+  delayMs: number;
+}
+
+export const defaultAvatarConfig: AvatarConfig = {
+  delayMs: 0
```
GODrums, line 8: Is `0ms` the best default value for us? I remember Shadcn recommending `600ms` to accommodate for slower connections, but unfortunately can't find any references for it right now. The official RadixUI example uses `600`, although the library has a default of `0` (https://www.radix-ui.com/primitives/docs/components/avatar).

GODrums (changes requested):
Great job, looks much cleaner than my attempt! My review was more of a learning experience of all of these (to me) new concepts, but it feels like a very natural implementation.

I am missing some testing abilities with the Storybook changes though (see my in-line comment) for the fallback or (potentially broken) image links.

webapp/src/app/ui/avatar/avatar.component.ts
FelixTJDietrich: Yeah, only the component selector is important. I think we could just omit `App` everywhere, when I used `ng g c avatar` it did not come with the prefix for the class. For consistency we'd need to adjust the other components, what do you think?

webapp/src/app/ui/avatar/avatar.stories.ts
FelixTJDietrich: I had some difficulties with that but I can try again, it assumes the props of `AvatarComponent`

webapp/src/app/ui/avatar/avatar-config.ts
FelixTJDietrich: Yeah I was also thinking the same, I also saw the 600 once. I think I will adjust it to prevent flickering

webapp/src/app/ui/avatar/avatar.stories.ts
FelixTJDietrich: I guess this works:
https://storybook.js.org/docs/writing-stories/args#args-can-modify-any-aspect-of-your-component

FelixTJDietrich: Done :)

webapp/src/app/ui/avatar/avatar.component.ts
GODrums: Agreed, let's remove it for all classes.

GODrums (approved):

PR #46, opened 2024-08-11
Update PR template

FelixTJDietrich:
### Description

I tried to improve the PR template a bit, we can improve it further.


FelixTJDietrich:
@GODrums and @milesha, feel free to give feedback, I think this structure might be better :)

GODrums (approved):

PR #50, opened 2024-08-11
Fix labeler workflow

FelixTJDietrich:
### Description

There is an error with the return. Hopefully this gets fixed now


FelixTJDietrich:
This can't be tested on a branch

PR #62, opened 2024-08-18
Add first GitHub entities and automatically fetch them for the leaderboard

GODrums:
### Motivation

To make our Leaderboard possible, we first have to fetch the required PRs / Comments data from Github's API and store them in the database, as well as make the data available through our own API.

### Description

- Adds `application-local.yml`-support for the Github PAT
- Adds the database structure for Github's API/Webhook schema
- Provides a sample runner to store Hephaestus`s data in the database
- Provides a Github API Client + GraphQL Client as fallback

Next up for the leaderboard (in other PRs):
- Webhook integrations (NATS)
- More endpoints depending on which data we need in our front-end

#### General

- [x] Documentation has been updated (if applicable)

#### Server (if applicable)


server/application-server/src/main/java/de/tum/in/www1/hephaestus/codereview/repository/Repository.java
```diff
@@ -0,0 +1,60 @@
...
+    @Column(name = "name_with_owner")
+    private String nameWithOwner;
+
+    @Column
+    private String description;
+
+    @Column
+    private String url;
+
+    @OneToMany(cascade = CascadeType.ALL, mappedBy = "repository", fetch = FetchType.EAGER)
```
FelixTJDietrich, line 49: We only want it to be eager if we return it via the rest api am I right? I think if we change to DTOs then the fetch type should be lazy?

server/application-server/src/main/java/de/tum/in/www1/hephaestus/codereview/repository/Repository.java
```diff
@@ -0,0 +1,60 @@
...
+     * Unique identifier for a Repository
+     */
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+
+    @Column(name = "github_id")
+    private Long githubId;
+
+    @Column
```
FelixTJDietrich, line 37: I think you can omit `@Column`, or you just always have it and fix the column name to some value.

server/application-server/src/main/java/de/tum/in/www1/hephaestus/codereview/repository/Repository.java
```diff
@@ -0,0 +1,60 @@
...
+    private String name;
+
+    @Column(name = "name_with_owner")
+    private String nameWithOwner;
+
+    @Column
+    private String description;
+
+    @Column
+    private String url;
```
FelixTJDietrich, line 47: What is this `url`? 
Is it the `clone_url` or the `html_url` or the REST API url? 

I think we should have two properties called:
`html_url` and `ssh_url`, the others are not important

server/application-server/src/main/java/de/tum/in/www1/hephaestus/codereview/repository/Repository.java
```diff
@@ -0,0 +1,60 @@
...
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+import lombok.Setter;
+
+@Entity
+@Table(name = "repositories")
+@Getter
+@Setter
+@NoArgsConstructor
+public class Repository {
```
FelixTJDietrich, line 25: Next to the properties you have listed I think the following are also interesting, potentially as a follow-up:
- `owner` / `owner_id` -> but we can do it as a followup, I think this would be a user object which is either an organization or a user AFAIK
- `pushed_at`
- `created_at`
- `updated_at`
- `homepage`
- `language`
- `default_branch`
- `open_issues_count`
- `stargazers_count`
- `watchers_count`
- `subscribers_count`
- `private`
- `archived`
- `disabled`
- `visibility`

server/application-server/src/main/java/de/tum/in/www1/hephaestus/codereview/repository/RepositoryController.java
```diff
@@ -0,0 +1,39 @@
...
+
+import java.util.List;
+
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.PathVariable;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RestController;
+
+@RestController
+@RequestMapping("/repository")
```
FelixTJDietrich, line 11: We can potentially just remove such code at the moment, if you don't want to maintain it. 

Right now we just need the schema and then have a controller for the leaderboard that we can fetch. No need to implement API endpoints that we might not use as of right now. But if you want to that's also fine

server/application-server/src/main/java/de/tum/in/www1/hephaestus/codereview/repository/RepositoryConverter.java
```diff
@@ -0,0 +1,31 @@
...
+import org.springframework.lang.Nullable;
+import org.springframework.stereotype.Service;
+
+@Service
+@ReadingConverter
+public class RepositoryConverter implements Converter<GHRepository, Repository> {
+
+    @Override
+    @Nullable
+    public Repository convert(@NonNull GHRepository source) {
```
FelixTJDietrich, line 19: Have you looked at MapStruct? Do we want to use it?

server/application-server/src/main/java/de/tum/in/www1/hephaestus/codereview/repository/RepositoryRepository.java
```diff
@@ -0,0 +1,14 @@
...
+import org.springframework.data.jpa.repository.JpaRepository;
+import org.springframework.data.jpa.repository.Query;
+
+import de.tum.in.www1.hephaestus.codereview.repository.Repository;
+
+@org.springframework.stereotype.Repository
+public interface RepositoryRepository
+                extends JpaRepository<Repository, Long> {
+
+        @Query("SELECT r FROM Repository r WHERE r.nameWithOwner = ?1")
```
FelixTJDietrich, line 12: This is again not necessart, JPA will automatically create this "simple" query

server/application-server/src/main/java/de/tum/in/www1/hephaestus/codereview/CodeReviewController.java
```diff
@@ -0,0 +1,30 @@
...
+@RequestMapping("/codereview")
+public class CodeReviewController {
+    private final CodeReviewService codeReviewService;
+
+    public CodeReviewController(CodeReviewService codeReviewService) {
+        this.codeReviewService = codeReviewService;
+    }
+
+    @PutMapping("/repository/{nameWithOwner}")
+    public Repository addRepository(@RequestParam String nameWithOwner) {
```
FelixTJDietrich, line 22: Is RequestParam correct?

server/application-server/src/main/java/de/tum/in/www1/hephaestus/codereview/CodeReviewController.java
```diff
@@ -0,0 +1,30 @@
...
+
+import org.springframework.web.bind.annotation.PutMapping;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestParam;
+import org.springframework.web.bind.annotation.RestController;
+
+import de.tum.in.www1.hephaestus.codereview.repository.Repository;
+
+@RestController
+@RequestMapping("/codereview")
```
FelixTJDietrich, line 13: I think this file can be removed in favor of a scheduled event:

```
package de.tum.in.www1.hephaestus.scheduler;

import de.tum.in.www1.hephaestus.repositorysync.GitHubRepositorySyncService;
import org.slf4j.Logger;
import org.slf4j.LoggerFactory;
import org.springframework.scheduling.annotation.Scheduled;
import org.springframework.stereotype.Component;

import java.io.IOException;

@Component
public class DailyGitHubDataSyncScheduler {

    private static final Logger logger = LoggerFactory.getLogger(DailyGitHubDataSyncScheduler.class);
    private final GitHubDataSyncService dataSyncService;

    public DailyGitHubDataSyncScheduler(GitHubDataSyncService dataSyncService) {
        this.dataSyncService = dataSyncService;
    }

    @Scheduled(cron = "0 0 3 * * ?") // Every day at 3 AM
    public void syncDataDaily() {
        logger.info("Starting daily GitHub data sync...");
        try {
            dataSyncService.syncData("ls1intum/hephaestus");
            logger.info("Daily GitHub data sync completed successfully.");
        } catch (IOException e) {
            logger.error("Error during GitHub data sync: ", e);
        }
    }
}
```

server/application-server/src/main/java/de/tum/in/www1/hephaestus/codereview/CodeReviewService.java
```diff
@@ -0,0 +1,212 @@
...
+import de.tum.in.www1.hephaestus.codereview.comment.CommentRepository;
+import de.tum.in.www1.hephaestus.codereview.pullrequest.Pullrequest;
+import de.tum.in.www1.hephaestus.codereview.pullrequest.PullrequestConverter;
+import de.tum.in.www1.hephaestus.codereview.pullrequest.PullrequestRepository;
+import de.tum.in.www1.hephaestus.codereview.repository.Repository;
+import de.tum.in.www1.hephaestus.codereview.repository.RepositoryConverter;
+import de.tum.in.www1.hephaestus.codereview.repository.RepositoryRepository;
+
+@Service
+public class CodeReviewService {
```
FelixTJDietrich, line 39: Call it `GitHubDataSyncService` and put it together with the schedular in a directory

server/application-server/src/main/java/de/tum/in/www1/hephaestus/codereview/CodeReviewService.java
```diff
@@ -0,0 +1,212 @@
...
+                logger.info("Hello from CodeReviewService!");
+
+                this.envConfig = envConfig;
+                this.repositoryRepository = repositoryRepository;
+                this.pullrequestRepository = pullrequestRepository;
+                this.commentRepository = commentRepository;
+                this.actorRepository = actorRepository;
+
+                RestClient restClient = RestClient.builder()
+                                .baseUrl("https://api.github.com/graphql")
```
FelixTJDietrich, line 66: The RestClient also uses the graphql baseUrl?

server/application-server/src/main/java/de/tum/in/www1/hephaestus/EnvConfig.java
```diff
@@ -0,0 +1,17 @@
+package de.tum.in.www1.hephaestus;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+@ConfigurationProperties(prefix = "environment")
+public class EnvConfig {
```
FelixTJDietrich, line 6: You can remove the file and use `@Value` in the service for example.

Articel: 
https://www.baeldung.com/spring-value-annotation

server/application-server/src/main/java/de/tum/in/www1/hephaestus/codereview/CodeReviewService.java
```diff
@@ -0,0 +1,212 @@
...
+                                .build();
+
+                String githubPat = this.envConfig.getGithubPat();
```
FelixTJDietrich, line 69: Use `@Value`, is also does not need to be a PAT? You could call it access token

server/application-server/src/main/java/de/tum/in/www1/hephaestus/codereview/CodeReviewService.java
```diff
@@ -0,0 +1,212 @@
...
+
+        private final HttpSyncGraphQlClient graphQlClient;
+
+        private GitHub github;
+
+        private final RepositoryRepository repositoryRepository;
+        private final PullrequestRepository pullrequestRepository;
+        private final CommentRepository commentRepository;
+        private final ActorRepository actorRepository;
+
```
FelixTJDietrich, line 51: Add something like:
```
@Value("${monitoring.repositories}")
private String[] repositoriesToMonitor;
```

And then have:
```
# application.yml

monitoring:
  repositories:
    - ls1intum/hephaestus
    - ls1intum/artemis
    - ls1intum/athena
    - ls1intum/pyris
```

server/application-server/src/main/java/de/tum/in/www1/hephaestus/codereview/CodeReviewController.java
FelixTJDietrich: In the applications.yml you could also have:
```
# application.yml

monitoring:
  repository-sync-cron: "0 0 3 * * ?"
```

And then:
```
@Scheduled(cron = "${monitoring.repository-sync-cron}")
```

server/application-server/src/main/java/de/tum/in/www1/hephaestus/Application.java
```diff
@@ -1,15 +1,31 @@
...
 public class Application {
 
 	public static void main(String[] args) {
 		SpringApplication.run(Application.class, args);
 	}
 
+	@Bean
+	CommandLineRunner commandLineRunner(CodeReviewService service) {
+		return args -> {
+			Repository repo = service.fetchRepository("ls1intum/hephaestus");
```
FelixTJDietrich, line 27: Fetching on startup makes sense to me but maybe this should also be configurable using `@Value`, same with the repositories to fetch.

server/application-server/src/main/java/de/tum/in/www1/hephaestus/codereview/CodeReviewService.java
```diff
@@ -0,0 +1,212 @@
...
+                return actor;
+        }
+
+        /**
+         * GraphQL implementation of fetching the hephaestus Github repository.
+         * 
+         * @see #fetchRepository(String)
+         * @return The hephaestus repository.
+         */
+        public Repository getHephaestusRepository() {
```
FelixTJDietrich, line 185: I think that could be removed, but I think you can keep the GraphQL client initialization as we might want to use it in the future. But I think we can also remove it completely for now until it is necessary again. Otherwise we would have unused code that we might also need to maintain.

FelixTJDietrich (changes requested):
I did not look at everything, but some comments also apply to the other files. Let's do this PR right and set the standard for future PRs :) Also as reference.

server/application-server/src/main/java/de/tum/in/www1/hephaestus/codereview/actor/Actor.java
```diff
@@ -0,0 +1,92 @@
...
+import jakarta.persistence.Entity;
+import jakarta.persistence.GeneratedValue;
+import jakarta.persistence.GenerationType;
+import jakarta.persistence.OneToMany;
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+import lombok.Setter;
+
+@Entity
+@Table(name = "actors")
```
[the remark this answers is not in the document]
GODrums: In my opinion both are fine, it's just a matter of convention. 
I've mostly used the plural in the past because makes the separation between the table and attributes that reference the table more clear.
I renamed it to the singular now to stay consistent with Artemis (bae8980).

server/application-server/src/main/java/de/tum/in/www1/hephaestus/codereview/actor/Actor.java
```diff
@@ -0,0 +1,92 @@
...
+@Getter
+@Setter
+@NoArgsConstructor
+@JsonInclude(JsonInclude.Include.NON_EMPTY)
+public class Actor {
```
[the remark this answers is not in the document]
GODrums: Agreed, the naming comes from Github's GraphQL API, which also calls it `Actor`:
image

I would be fine with renaming it. Do we plan on using different tables for Github and Gitlab? Then `GHUser` and `GLUser` would be perfect. Otherwise maybe `ExternalUser`?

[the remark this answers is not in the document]
FelixTJDietrich: I think we will have one table for each entity and try our best to union it

server/application-server/src/main/java/de/tum/in/www1/hephaestus/codereview/actor/Actor.java
```diff
@@ -0,0 +1,92 @@
...
+    /**
+     * Unique identifier for a User entity.
```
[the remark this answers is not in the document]
GODrums: `@GeneratedValue(strategy = GenerationType.IDENTITY)` means we generate this id ourselves on table-insert.

For users I thought it might make sense to leave out Github's ID since `String login` should be unique anyways.

Would you like me to still add the `github_id`?

server/application-server/src/main/java/de/tum/in/www1/hephaestus/codereview/actor/Actor.java
```diff
@@ -0,0 +1,92 @@
...
+
+    @Column
+    private String email;
+
+    /**
+     * URL of the User entity.
+     * This field is mandatory.
+     */
+    @Column(nullable = false)
+    private String url;
```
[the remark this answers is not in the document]
GODrums: yes, this is exactly how the url is meant to be, i.e. the "profile url".
For example, the stored `url` for my account is https://github.com/GODrums, see the response from the endpoint:
image

server/application-server/src/main/java/de/tum/in/www1/hephaestus/codereview/actor/Actor.java
```diff
@@ -0,0 +1,92 @@
...
+    public void addComment(Comment comment) {
+        if (comments == null) {
+            comments = new ArrayList<>();
+        }
+        if (!comments.contains(comment)) {
+            comments.add(comment);
+        }
+    }
+
+    public void addPullrequest(Pullrequest pullrequest) {
```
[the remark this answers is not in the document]
GODrums: Only for this method or also for the `Pullrequest`-Entity? I think we should consistent, so if we rename it here, we should also rename the table + attributes.

server/application-server/src/main/java/de/tum/in/www1/hephaestus/codereview/actor/Actor.java
```diff
@@ -0,0 +1,92 @@
...
+        if (pullrequests == null) {
+            pullrequests = new ArrayList<>();
+        }
+        if (!pullrequests.contains(pullrequest)) {
+            pullrequests.add(pullrequest);
+        }
+    }
+
+    @Override
+    public String toString() {
```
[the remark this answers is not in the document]
GODrums: Great find in Lombok! It is kind of sad, that we cannot print these relations at all though, as I believe these provide great additional value (especially when observing behavior).

server/application-server/src/main/java/de/tum/in/www1/hephaestus/codereview/actor/Actor.java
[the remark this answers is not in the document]
GODrums: For now I wanted to stick to the most important attributes, I personally would just add everthing "on-demand" to not spam our database. Would you agree with that approach?

Totally forgot about about the `name` as attribute though, just added it in 05ed344.

We don't really need the `avatar_url` explicitly as it as available on Github via `{profile_url}.png`

server/application-server/src/main/java/de/tum/in/www1/hephaestus/codereview/repository/Repository.java
GODrums: Currently it refers to the Website/HTML Url. I guess an interesting question here is if we really need any of these URLs. All of them are pretty easy to generate:
- HTML URL: `https://github.com/{nameWithOwner}`
- Clone HTTPS URL: `https://github.com/{nameWithOwner}.git`
- Clone SSH URL: `git@github.com:{nameWithOwner}.git`

Do you think we should store any of them?

server/application-server/src/main/java/de/tum/in/www1/hephaestus/codereview/user/GHUser.java
```diff
@@ -0,0 +1,85 @@
...
+import lombok.Setter;
+import lombok.ToString;
+
+@Entity
+@Table(name = "gh_user")
+@Getter
+@Setter
+@NoArgsConstructor
+@ToString
+public class GHUser {
```
FelixTJDietrich, line 32: Does it make sense to generalize it? GitLab has also roughly the same fields, same with "Repository" on GH and "Project" on GL.

We could also add our own user attributes here if necessary. I think it makes sense to keep it general

server/application-server/src/main/java/de/tum/in/www1/hephaestus/codereview/actor/Actor.java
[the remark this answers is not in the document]
FelixTJDietrich: I think we can use `id` everywhere for now instead of `github_id`. In the future we would then just add a `provider_id` field in addition. 

GitLab uses also an integer for the `id` (Bitbucket as far as I can see too) and others are not interesting for the moment. I think this would be the clearest solution.

server/application-server/src/main/java/de/tum/in/www1/hephaestus/codereview/actor/Actor.java
```diff
@@ -0,0 +1,92 @@
...
+
+    /**
+     * The Pullrequests of the User entity.
+     */
+    @OneToMany(mappedBy = "author")
+    @JsonIgnore
+    private List<Pullrequest> pullrequests;
+
+    /**
+     * The Comments of the User entity.
```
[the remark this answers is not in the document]
FelixTJDietrich: I meant that maybe some comments are obvious from the from the code and don't need further explanation that repeats it.

For instance:
```
/**
 * Unique identifier for a Comment entity.
 */
@Id
@GeneratedValue(strategy = GenerationType.IDENTITY)
private Long id;
```

I know that this is a field of the Comment entity, it's tagged with `@Id` so I know it is a identifier and also unique.

Another example:
```
/**
 * Body of the Comment entity.
 * This field is mandatory.
 */
@NonNull
private String body;
```

I think in such cases we can omit the comment because it does not say more than the code itself. What do you think?

server/application-server/src/main/java/de/tum/in/www1/hephaestus/codereview/actor/Actor.java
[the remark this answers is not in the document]
FelixTJDietrich: I would have just stored the `avatar_url` it might be different for GitLab and now the client needs to know that the avatar is located at `{profile_url}.png`. We don't have to be that frugal with the data we are storing.

server/application-server/src/main/java/de/tum/in/www1/hephaestus/codereview/repository/Repository.java
FelixTJDietrich: Alright, no I think it is fine the way you are doing it. We don't really need the others.

server/application-server/src/main/java/de/tum/in/www1/hephaestus/codereview/actor/Actor.java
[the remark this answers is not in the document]
FelixTJDietrich: But then you'd also always need to fetch those relationships 🤔

server/application-server/src/main/java/de/tum/in/www1/hephaestus/codereview/pullrequest/Pullrequest.java
```diff
@@ -0,0 +1,110 @@
...
+import lombok.ToString;
+
+@Entity
+@Table(name = "pull_request")
+@Getter
+@Setter
+@NoArgsConstructor
+@JsonInclude(JsonInclude.Include.NON_EMPTY)
+@ToString
+public class PullRequest {
```
FelixTJDietrich, line 37: I just saw that in the RestAPI `GHPullRequest` inherits from `GHIssue`. Maybe we can also do some inheritance. 

I also saw this article about it: https://www.baeldung.com/hibernate-inheritance

Maybe we can have a `@MappedSuperclass` `BaseGitServiceEntity` that we can extend everywhere adding the `id`, `created_at`, and `updated_at` there and in the future also `provider_id`.

server/application-server/src/main/java/de/tum/in/www1/hephaestus/codereview/user/GHUser.java
GODrums: So you would suggest a super-class (e.g. `ProviderUser`) and then sub-classes for GHUser and GLUser? Or just one class for all of them?
Also we might want to leave the `class User` open for when we create our own login/account system.

FelixTJDietrich: Just use one class for them and call also call it user for now. GitHub and GitLab use a similar authentication, me might add the auth stuff to the same table since every user might also login with Hephaestus.

server/application-server/src/main/java/de/tum/in/www1/hephaestus/codereview/comment/IssueCommentConverter.java
```diff
@@ -0,0 +1,31 @@
+package de.tum.in.www1.hephaestus.codereview.comment;
+
+import java.io.IOException;
+
+import org.kohsuke.github.GHIssueComment;
+import org.springframework.core.convert.converter.Converter;
+import org.springframework.lang.NonNull;
+
+public class IssueCommentConverter implements Converter<GHIssueComment, IssueComment> {
```
FelixTJDietrich, line 9: To avoid any errors and more boilerplate code we could potentially do an abstract base class (code by ChatGPT):
```
package de.tum.in.www1.hephaestus.codereview.converter;

import de.tum.in.www1.hephaestus.codereview.base.BaseGitServiceEntity;
import org.kohsuke.github.GHObject;
import org.springframework.core.convert.converter.Converter;
import org.slf4j.Logger;
import org.slf4j.LoggerFactory;
import java.io.IOException;
import java.time.OffsetDateTime;
import java.time.ZoneOffset;
import java.util.Date;

public abstract class BaseGitServiceEntityConverter<S extends GHObject, T extends BaseGitServiceEntity> implements Converter<S, T> {

    private static final Logger logger = LoggerFactory.getLogger(BaseGitServiceEntityConverter.class);

    protected void convertBaseFields(S source, T target) {
        if (source == null || target == null) {
            throw new IllegalArgumentException("Source and target must not be null");
        }

        // Map common fields
        target.setId(source.getId());

        try {
            target.setCreatedAt(convertToOffsetDateTime(source.getCreatedAt()));
        } catch (IOException e) {
            logger.error("Failed to convert createdAt field for source {}: {}", source.getId(), e.getMessage());
            target.setCreatedAt(null);
        }

        try {
            target.setUpdatedAt(convertToOffsetDateTime(source.getUpdatedAt()));
        } catch (IOException e) {
            logger.error("Failed to convert updatedAt field for source {}: {}", source.getId(), e.getMessage());
            target.setUpdatedAt(null);
        }
    }

    private OffsetDateTime convertToOffsetDateTime(Date date) {
        return date != null ? date.toInstant().atOffset(ZoneOffset.UTC) : null;
    }
}
```

And then:
```
public class IssueCommentConverter extends BaseGitServiceEntityConverter<GHIssueComment, IssueComment> {

    @Override
    public IssueComment convert(@NonNull GHIssueComment source) {
        IssueComment target = new IssueComment();
        super.convertBaseFields(source, target); 
        ...
```

server/application-server/src/main/java/de/tum/in/www1/hephaestus/codereview/comment/IssueCommentConverter.java
```diff
@@ -0,0 +1,31 @@
...
+
+    @Override
+    public IssueComment convert(@NonNull GHIssueComment source) {
+        IssueComment comment = new IssueComment();
```
FelixTJDietrich, line 13: I think you missed the id?

server/application-server/src/main/java/de/tum/in/www1/hephaestus/codereview/comment/IssueCommentConverter.java
```diff
@@ -0,0 +1,31 @@
...
+            comment.setCreatedAt(null);
+        }
+        try {
+            comment.setUpdatedAt(source.getUpdatedAt().toString());
+        } catch (IOException e) {
+            comment.setUpdatedAt(null);
+        }
+        // set preliminary values to be filled in later
+        comment.setPullrequest(null);
+        comment.setAuthor(null);
```
FelixTJDietrich, line 27: Those are `null` by default?

server/application-server/src/main/java/de/tum/in/www1/hephaestus/codereview/comment/IssueCommentDTO.java
```diff
@@ -0,0 +1,14 @@
+package de.tum.in.www1.hephaestus.codereview.comment;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+
+import de.tum.in.www1.hephaestus.codereview.pullrequest.PullRequestDTO;
+import de.tum.in.www1.hephaestus.codereview.user.UserDTO;
+
+@JsonInclude(JsonInclude.Include.NON_EMPTY)
+public record IssueCommentDTO(String body, String createdAt, String updatedAt, UserDTO author,
```
FelixTJDietrich, line 9: Is there a reason for not including the id?

server/application-server/src/main/java/de/tum/in/www1/hephaestus/codereview/pullrequest/GHIssueState.java
```diff
@@ -0,0 +1,5 @@
+package de.tum.in.www1.hephaestus.codereview.pullrequest;
+
+public enum GHIssueState {
+    ALL, CLOSED, OPEN
```
FelixTJDietrich, line 4: What is `ALL`? GitHub says: "State of the issue; either 'open' or 'closed'"

server/application-server/src/main/java/de/tum/in/www1/hephaestus/codereview/pullrequest/GHIssueState.java
```diff
@@ -0,0 +1,5 @@
+package de.tum.in.www1.hephaestus.codereview.pullrequest;
+
+public enum GHIssueState {
```
FelixTJDietrich, line 3: You can omit the `GH` prefix here

server/application-server/src/main/java/de/tum/in/www1/hephaestus/codereview/pullrequest/Pullrequest.java
```diff
@@ -0,0 +1,64 @@
...
+import lombok.ToString;
+
+@Entity
+@Table(name = "pull_request")
+@Getter
+@Setter
+@NoArgsConstructor
+@JsonInclude(JsonInclude.Include.NON_EMPTY)
+@ToString(callSuper = true)
+public class PullRequest extends BaseGitServiceEntity {
```
FelixTJDietrich, line 34: I think you need to rename the files starting with `PullRequest`, I'm getting an error here or is this on my side?

server/application-server/src/main/java/de/tum/in/www1/hephaestus/codereview/pullrequest/PullrequestController.java
```diff
@@ -0,0 +1,28 @@
...
+@RequestMapping("/pr")
+public class PullRequestController {
+    private final PullRequestService pullrequestService;
+
+    public PullRequestController(PullRequestService pullrequestService) {
+        this.pullrequestService = pullrequestService;
+    }
+
+    @GetMapping("/{id}")
+    public PullRequest getPullrequest(@PathVariable Long id) {
```
FelixTJDietrich, line 20: Can you search and replace `Pullrequest` with `PullRequest` for all files

server/application-server/src/main/java/de/tum/in/www1/hephaestus/codereview/pullrequest/PullrequestController.java
```diff
@@ -0,0 +1,28 @@
...
+
+import java.util.List;
+
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.PathVariable;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RestController;
+
+@RestController
+@RequestMapping("/pr")
```
FelixTJDietrich, line 11: Github names this mapping `pulls` maybe name it `pullrequests` or `pullrequest`? But if you want to keep `pr` that's also fine

server/application-server/src/main/java/de/tum/in/www1/hephaestus/codereview/pullrequest/GHIssueState.java
FelixTJDietrich: I think it is more for the filtering case

server/application-server/src/main/java/de/tum/in/www1/hephaestus/codereview/pullrequest/PullrequestConverter.java
```diff
@@ -0,0 +1,52 @@
...
+            pullrequest.setUpdatedAt(null);
+        }
+        pullrequest.setMergedAt(source.getMergedAt() != null ? source.getMergedAt().toString() : null);
+        // set preliminary values to be filled in later
+        pullrequest.setAuthor(null);
+        pullrequest.setRepository(null);
+        return pullrequest;
+    }
+
+    private GHIssueState convertState(org.kohsuke.github.GHIssueState state) {
```
FelixTJDietrich, line 39: I think here you can actually only get `OPEN` or `CLOSED`

server/application-server/src/main/java/de/tum/in/www1/hephaestus/codereview/pullrequest/PullrequestRepository.java
```diff
@@ -0,0 +1,15 @@
...
+
+import org.springframework.data.jpa.repository.JpaRepository;
+import org.springframework.data.jpa.repository.Query;
+import org.springframework.stereotype.Repository;
+
+@Repository
+public interface PullRequestRepository extends JpaRepository<PullRequest, Long> {
+
+    @Query("SELECT p FROM PullRequest p, User a WHERE p.author = a AND a.login = ?1")
+    List<PullRequest> findByAuthor(String authorLogin);
```
FelixTJDietrich, line 13: You can write it like this and use a `Set`:
```suggestion
    Set<PullRequest> findByAuthor_Login(String authorLogin);
```

server/application-server/src/main/java/de/tum/in/www1/hephaestus/codereview/repository/Repository.java
```diff
@@ -0,0 +1,43 @@
...
+@Table(name = "repository")
+@Getter
+@Setter
+@NoArgsConstructor
+@ToString(callSuper = true)
+public class Repository extends BaseGitServiceEntity {
+    @NonNull
+    private String name;
+
+    @Column(name = "name_with_owner")
```
FelixTJDietrich, line 31: This can be removed? 
https://www.baeldung.com/hibernate-field-naming-spring-boot

server/application-server/src/main/java/de/tum/in/www1/hephaestus/codereview/user/User.java
```diff
@@ -0,0 +1,72 @@
...
+import jakarta.persistence.Column;
+import jakarta.persistence.Entity;
+import jakarta.persistence.OneToMany;
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+import lombok.Setter;
+import lombok.ToString;
+
+@Entity
+@Table(name = "user", schema = "public")
```
FelixTJDietrich, line 23: What is `schema = "public"` can we remove it?

server/application-server/src/main/java/de/tum/in/www1/hephaestus/codereview/user/UserRepository.java
```diff
@@ -0,0 +1,29 @@
...
+import org.springframework.data.jpa.repository.Query;
+import org.springframework.data.repository.query.Param;
+
+public interface UserRepository extends JpaRepository<User, Long> {
+
+    @Query("SELECT u FROM User u WHERE u.login = :login")
+    Optional<User> findUser(@Param("login") String login);
+
+    @Query("""
+                SELECT new de.tum.in.www1.hephaestus.codereview.user.UserDTO(u.login, u.email, u.name, u.url)
```
FelixTJDietrich, line 15: Apparently we can use the `ClassImportIntegrator` utility from [Hypersistence Utils project](https://github.com/vladmihalcea/hypersistence-utils) to avoid having to write fully-qualified names of Java classes.

```
public class ClassImportIntegratorIntegratorProvider implements IntegratorProvider {
 
    @Override
    public List<Integrator> getIntegrators() {
        return List.of(
            new ClassImportIntegrator(
                List.of(
                    PostDTO.class
                )
            )
        );
    }
}
```

And then in `application.yaml`:
```
spring:
  jpa:
    properties:
      hibernate:
        integrator_provider=your.awesome.app.ClassImportIntegratorIntegratorProvider
```

Can you try that?
Here is the blog post where I found it:
https://vladmihalcea.com/dto-projection-jpa-query/

server/application-server/src/main/java/de/tum/in/www1/hephaestus/codereview/user/UserRepository.java
```diff
@@ -0,0 +1,29 @@
...
+                FROM User u
+                WHERE u.login = :login
+            """)
+    Optional<UserDTO> findByLogin(@Param("login") String login);
```
FelixTJDietrich, line 19: Bonus to the comment above that we should anyways do, can you try removing the `@Query` and see if it still works? I think since the `UserDTO` is a subset of `User` without type conversions this should automatically be handled by hibernate.

server/application-server/src/main/java/de/tum/in/www1/hephaestus/codereview/user/UserRepository.java
```diff
@@ -0,0 +1,29 @@
...
+
+    // @Query("""
+    // SELECT new de.tum.in.www1.hephaestus.codereview.user.UserDTO(u.login,
+    // u.email, u.name, u.url, u.pullRequests, u.comments)
+    // FROM User u
+    // WHERE u.login = :login
+    // """)
+    // Optional<UserDTO> findByLoginWithPullRequestsAndComments(@Param("login")
+    // String login);
```
FelixTJDietrich, line 28: Remove comment

server/application-server/pom.xml
```diff
@@ -29,6 +29,7 @@
 	</scm>
 	<properties>
 		<java.version>21</java.version>
+		<app.profiles>local,dev</app.profiles>
```
FelixTJDietrich, line 32: We also have `prod`, am I right? At least that is what I'm using in the docker compose file `SPRING_PROFILES_ACTIVE=prod`

FelixTJDietrich (changes requested):

server/application-server/src/main/java/de/tum/in/www1/hephaestus/Application.java
FelixTJDietrich: Example how you could move it out of the `CommandLineRunner` and use `Value`:
```
@Value("${monitoring.runOnStartup:true}")
private boolean runOnStartup;

@PostConstruct
public void onStartup() {
    if (runOnStartup) {
        doWork();
    }
}
```

I think this would be preferred

server/application-server/src/main/java/de/tum/in/www1/hephaestus/codereview/comment/IssueComment.java
```diff
@@ -0,0 +1,49 @@
...
+    @ToString.Exclude
+    private User author;
+
+    @ManyToOne(optional = false)
+    @JoinColumn(name = "pullrequest_id", referencedColumnName = "id")
+    @JsonIgnore
+    @ToString.Exclude
+    private PullRequest pullrequest;
+
+    public IssueComment(String body, String createdAt, String updatedAt) {
```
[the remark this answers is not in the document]
GODrums: The ID is auto-generated, right? So we don't need it in the constructor.
I will give `@AllArgsConstructor` a try, didn't know that existed.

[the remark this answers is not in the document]
FelixTJDietrich: I see, actually, I think we need to remove `@GeneratedValue(strategy = GenerationType.IDENTITY)` for all GitServiceEntities, now I understand it. We would be using the ID that comes from GitHub and not generate it on our own.

server/application-server/src/main/java/de/tum/in/www1/hephaestus/codereview/pullrequest/Pullrequest.java
GODrums: Seems like you are right, Git did not push the rename. For me in VS Code everything works and is correctly renamed.
I can try it in a different environment.

server/application-server/src/main/java/de/tum/in/www1/hephaestus/codereview/pullrequest/PullrequestConverter.java
FelixTJDietrich: I know, but github's REST API docs only say open and closed 🤔

server/application-server/src/main/java/de/tum/in/www1/hephaestus/codereview/pullrequest/PullrequestRepository.java
GODrums: That's very interesting! Is there maybe a good "documentation" for the naming I could learn from?

FelixTJDietrich: https://docs.spring.io/spring-data/jpa/reference/repositories/query-methods-details.html

i will also read through it, maybe we don't even need the `_`

server/application-server/src/main/java/de/tum/in/www1/hephaestus/codereview/user/User.java
GODrums: The table name `user` is a reserved keyword in Postgres. That's why we have to specify the schema, see the [PostgreSQL Docs](https://www.postgresql.org/docs/8.1/sql-keywords-appendix.html) or this [Stackoverflow post](https://stackoverflow.com/questions/45782327/org-postgresql-util-psqlexception-error-column-user0-id-does-not-exist-hibe)

server/application-server/pom.xml
GODrums: Yes, indeed. Via the command I always had trouble selecting a profile though, so adding it here would auto-overwrite all my values with the prod-placeholders.

FelixTJDietrich: I see, on Artemis the default profile is also `dev` 👍

server/application-server/src/main/java/de/tum/in/www1/hephaestus/codereview/user/UserRepository.java
GODrums: In theory it would work, but only for the full arg constructors as it seems. Since here we want to use the smaller constructor, it doesn't work (just tested it).

server/application-server/src/main/java/de/tum/in/www1/hephaestus/codereview/comment/IssueComment.java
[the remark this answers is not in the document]
GODrums: Oh, so you want to use Github's ID instead? I thought we would later on just add a `provider_id` variable. I am personally not sure if using Github's ID is a good idea since there is no guarantee these are unique across providers.

server/application-server/src/main/java/de/tum/in/www1/hephaestus/codereview/user/UserRepository.java
FelixTJDietrich: Good to know, thanks!

server/application-server/src/main/java/de/tum/in/www1/hephaestus/codereview/comment/IssueComment.java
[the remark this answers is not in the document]
GODrums: Thank you for clarifying! That makes way more sense.

I also gave `@AllArgsConstructor` a shot. Unfortunately, it seems to be a known issue that it does not allow to initialize attributes of superclasses, so we will have to create the constructors ourselves. See [lombok#2624](https://github.com/projectlombok/lombok/issues/2646).

server/application-server/src/main/java/de/tum/in/www1/hephaestus/codereview/comment/IssueComment.java
```diff
@@ -0,0 +1,49 @@
...
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+import lombok.Setter;
+import lombok.ToString;
+
+@Entity
+@Table(name = "issue_comment")
+@Getter
+@Setter
+@NoArgsConstructor
```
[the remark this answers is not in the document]
GODrums: Unfortunately, hibernate forces you to have a default constructor (i.e. no args). See [this post](https://stackoverflow.com/questions/44088360/org-hibernate-instantiationexception-no-default-constructor-for-entity-princ)

server/application-server/src/main/java/de/tum/in/www1/hephaestus/codereview/repository/Repository.java
GODrums: I added a few of these that felt like they could be useful for the leaderboard (commit f9af54a)

server/application-server/src/main/java/de/tum/in/www1/hephaestus/codereview/user/UserController.java
```diff
@@ -0,0 +1,30 @@
...
+
+    public UserController(UserService actorService) {
+        this.userService = actorService;
+    }
+
+    @GetMapping("/{login}")
+    public UserDTO getUser(@PathVariable String login) {
+        Optional<UserDTO> user = userService.getUserDTO(login);
+        if (user.isEmpty()) {
+            throw new EntityNotFoundException("Actor with login " + login + " not found!");
```
FelixTJDietrich, line 25: Can you look into `ResponseEntity`? 
On Artemis they just do `ResponseEntity.notFound().build()` and also `ResponseEntity.ok(user.get())` for the ok response.

FelixTJDietrich (commented):
Some more final comments

server/application-server/src/main/resources/application.yml
```diff
@@ -1,26 +1,40 @@
...
+            password: admin
 
 springdoc:
-  default-produces-media-type: application/json
+    default-produces-media-type: application/json
+
+monitoring:
+    runOnStartup: true
+    repository-sync-cron: "0 0 3 * * ?"
+    repositories: ls1intum/hephaestus, ls1intum/pyris
```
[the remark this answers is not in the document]
GODrums: > Can't you put it on multiple lines like
> 
> ```
> repositoris:
>   - ls1intum/hephaestus
>   - ls1intum/pyris
> ```

Unfortunately, `@Value` seems to be not able to parse this multi-line list format. When debugging this, I found [this post](https://stackoverflow.com/questions/26699385/spring-boot-yaml-configuration-for-a-list-of-strings) with a few detailed explanation and options.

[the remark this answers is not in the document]
FelixTJDietrich: Alright 👍 then we have to keep it like this :/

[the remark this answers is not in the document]
FelixTJDietrich: But I guess we can still do it multiline 

image

FelixTJDietrich:
I think the code looks good now, maybe you can take a quick look at `ResponseEntity` and the returned `NotFoundException`. Also you need to update the angular API client library, there is a command that you need to run in `package.json`. Afterwards we can merge this PR :)

server/application-server/src/main/java/de/tum/in/www1/hephaestus/codereview/comment/IssueCommentConverter.java
```diff
@@ -2,9 +2,11 @@
 
 import org.kohsuke.github.GHIssueComment;
 import org.springframework.lang.NonNull;
+import org.springframework.stereotype.Service;
 
 import de.tum.in.www1.hephaestus.codereview.base.BaseGitServiceEntityConverter;
 
+@Service
```
FelixTJDietrich, line 9: I guess `@Component` for the converters is more fitting.

FelixTJDietrich: You could also move the annotation to the `BaseGitServiceEntityConverter`

FelixTJDietrich (commented):
I have only a minor comment, the rest looks fine. You still need to make the checks pass.

server/application-server/src/main/java/de/tum/in/www1/hephaestus/codereview/comment/IssueCommentConverter.java
GODrums: > You could also move the annotation to the `BaseGitServiceEntityConverter`

It tried this, but fortunately it isn't enough for the compiler to auto-inject all subclasses. Changed it to `@Component` now though.

FelixTJDietrich (approved):
Code LGTM now, thank you!

PR #78, opened 2024-09-08
Leaderboard component and homepage

GODrums:
### Motivation

As our first gamfication component, we want to create a rework of the activity leaderboard in Artemis.

### Pre-requisits:
- https://github.com/ls1intum/Hephaestus/pull/76
- https://github.com/ls1intum/Hephaestus/pull/80

### Description

This PR adds a simple Artemis leaderboard to the homepage and creates components with corresponding Storybook-Stories.
In the future, it can easily extended with `@tanstack/angular-table` for filters, sorting, pagination.

image

#### General

#### Client (if applicable)

- [x] Added Storybook stories for new components
- [x] Components follow design system guidelines (if applicable)


server/application-server/src/main/java/de/tum/in/www1/hephaestus/codereview/user/UserType.java
```diff
@@ -0,0 +1,5 @@
+package de.tum.in.www1.hephaestus.codereview.user;
+
+public enum UserType {
```
FelixTJDietrich, line 3: Technically I think the user type can also be organization. For now this is enough :)

webapp/src/app/components/leaderboard/leaderboard.component.ts
```diff
@@ -0,0 +1,153 @@
...
+import { TableCaptionDirective } from 'app/ui/table/table-caption.directive';
+import { TableCellDirective } from 'app/ui/table/table-cell.directive';
+import { TableFooterDirective } from 'app/ui/table/table-footer.directive';
+import { TableHeadDirective } from 'app/ui/table/table-head.directive';
+import { TableHeaderDirective } from 'app/ui/table/table-header.directive';
+import { TableRowDirective } from 'app/ui/table/table-row.directive';
+import { TableComponent } from 'app/ui/table/table.component';
+import { lastValueFrom } from 'rxjs';
+
+const defaultData: LeaderboardEntry[] = [
```
FelixTJDietrich, line 17: Hmm, I think the way to do it would be mocking the service for the storybook story somehow.

FelixTJDietrich: This data should be part of the story though

webapp/src/app/components/leaderboard/leaderboard.component.ts
```diff
@@ -0,0 +1,153 @@
...
+
+  query = injectQuery(() => ({
+    queryKey: ['leaderboard'],
+    queryFn: async () => lastValueFrom(this.leaderboardService.getLeaderboard()),
+    gcTime: Infinity
+  }));
+
+  leaderboard = computed(() => {
+    let data = this.query.data() ?? defaultData;
+    data = data.filter((entry) => entry.type === LeaderboardEntry.TypeEnum.User).sort(sortByScore);
```
FelixTJDietrich, line 150: I think this logic could be moved to the server side. Sorting and filtering out the bot users

webapp/src/app/components/leaderboard/leaderboard.component.html
```diff
@@ -0,0 +1,55 @@
...
+                  <img src="{{ entry.avatarUrl }}" width="24" height="24" />
+                  <span class="text-muted-foreground">{{ entry.name }}</span>
+                </a>
+              </td>
+              <td appTableCell>{{ entry.score }}</td>
+              <td appTableCell class="flex items-center gap-4" title="Changes Requested">
+                <div class="flex items-center gap-2">
+                  <app-icon-pull-request-changes-requested />
+                  {{ entry.changesRequested }}
+                </div>
```
FelixTJDietrich, line 33: Have you checked how it looks when you remove the `0` value activities with `@if`? I think it would look good

webapp/src/app/components/leaderboard/leaderboard.stories.ts
```diff
@@ -0,0 +1,18 @@
+import { type Meta, type StoryObj } from '@storybook/angular';
+import { LeaderboardComponent } from './leaderboard.component';
+
+const meta: Meta<LeaderboardComponent> = {
+  title: 'Components/Leaderboard',
+  component: LeaderboardComponent,
+  tags: ['autodocs']
```
FelixTJDietrich, line 7: I think we would somehow inject a provider for a mocked `LeaderboardService` that returns the mock data here. Alternatively, we would dumb down the leaderboard component into for example `LeaderboardComponent` (smart with service) and `LeaderboardTableComponent` (dumb no service).

FelixTJDietrich (approved):
Generally looks good :) We can also do the improvements in a follow-up

webapp/src/app/ui/icons/PullRequestApprovedIcon.component.ts
FelixTJDietrich, line 1: For the icon components we would also want stories. Maybe just a big story with all custom icons 🤔 But we can also do this in a follow-up

webapp/src/app/components/leaderboard/leaderboard.component.html
```diff
@@ -0,0 +1,55 @@
+<div class="flex flex-col gap-4 py-8">
+  <h1 class="text-3xl font-bold">Artemis Leaderboard</h1>
```
FelixTJDietrich, line 2: Maybe just have a leaderboard page component that is the smart component and separate out the table, I think this is the way. 

We can skip having a story for the smart components for now, there we would need to mock the services in the stories.

webapp/src/app/components/leaderboard/leaderboard.component.html
```diff
@@ -0,0 +1,55 @@
...
+                </div>
+                <div class="flex items-center gap-2" title="Comments">
+                  <app-icon-pull-request-comment />
+                  {{ entry.comments }}
+                </div>
+              </td>
+            </tr>
+          }
+        </tbody>
+        <tfoot appTableFooter>
```
FelixTJDietrich, line 46: I think we can remove the footer row actually 🤔

webapp/src/app/components/leaderboard/leaderboard.component.html
```diff
@@ -0,0 +1,55 @@
...
+  } @else if (query.error()) {
+    <span class="text-destructive">An error has occurred</span>
+  }
+
+  @if (leaderboard(); as leaderboard) {
+    <div class="px-4">
+      <app-table>
+        <thead appTableHeader>
+          <tr appTableRow>
+            <th appTableHead>Contributor</th>
```
FelixTJDietrich, line 14: A column for placement would be good I guess

server/application-server/src/main/java/de/tum/in/www1/hephaestus/leaderboard/LeaderboardService.java
```diff
@@ -27,10 +28,10 @@ public List<LeaderboardEntry> createLeaderboard() {
         logger.info("Found " + users.size() + " users");
```
FelixTJDietrich, line 28: We should also show users that are in the development team but did not do any reviewing

webapp/src/app/components/leaderboard/leaderboard.component.ts
GODrums: We can probably just delete this data set as it should only be for testing purposes. And then only display the leaderboard if we are able to fetch data from the server. The story currently uses its own testing data set.


Would that solve the issue for you?

FelixTJDietrich: I mean the story fails to fetch the data and then falls back to this data, either remove it here or move it to the story.

webapp/src/app/components/leaderboard/leaderboard.component.ts
GODrums: I was under the assumption that in one of the next versions we want to implement custom table filters / sorting for the user. So that for example a user is able to display BOT-accounts or sort after a different criterium.
Then we need to have filtering/sorting logic in the client anyways I guess?

FelixTJDietrich: I don't think that code rabbit is interesting to see on the leaderboard we don't need a filter function for just hiding that. I think we can just generally exclude it.

Sorting does also not make sense to change for a leaderboard, or at least is not essential at the moment. 

The next step would be filtering in regard to time frame IMO, so you can see the leaderboard over the last week, month, custom or so.

server/application-server/src/main/java/de/tum/in/www1/hephaestus/leaderboard/LeaderboardService.java
FelixTJDietrich: On the preview deployment it they are visible 🤔 I guess I was mistaken

webapp/src/app/main/main.component.ts
```diff
@@ -1,11 +1,10 @@
 import { Component } from '@angular/core';
-import { CounterComponent } from 'app/example/counter/counter.component';
-import { HelloComponent } from 'app/example/hello/hello.component';
+import { LeaderboardComponent } from 'app/components/leaderboard/leaderboard.component';
 
 @Component({
   selector: 'app-main',
   standalone: true,
-  imports: [CounterComponent, HelloComponent],
```
FelixTJDietrich, line 8: I removed the unused code in https://github.com/ls1intum/Hephaestus/pull/84

webapp/src/app/components/leaderboard/leaderboard.component.ts
GODrums: I personally thought about either timeframe or repository filters (as soon as we start supporting multiple repos).

But you are right actually, it indeed might make more sense to move the logic to the server. Will change that.

FelixTJDietrich: True repository filters also, we can just have request parameters for the filtering and let the server take care of it. Angular/react query works really well with that.

webapp/src/app/components/leaderboard/leaderboard.component.html
GODrums: This is how it would look for the past 7 days:
image
It will be much more filled when we extend the timeframe to 30days of course, but I also like it more.

FelixTJDietrich: Looks really good :) 
On that note, why is Kaan having score 68 without having any activity? Are we missing something?

webapp/src/app/components/leaderboard/leaderboard.component.html
GODrums: Right now the leaderboard is already basically just a wrapper for the `table`-component with the required logic it feels like.
I guess we could also just move the heading to the `main`-component to make the separation more clear?
Is there an advantage to separate components I'm missing here?

webapp/src/app/components/leaderboard/leaderboard.component.html
GODrums: Absolutely, right now it provides no real value.
Thinking about it, we could also use it to sum up all of the activity in the leaderboard.

Which option would you prefer?

webapp/src/app/components/leaderboard/leaderboard.component.html
FelixTJDietrich: I'm still thinking about it 🤔 Okay maybe taking a step back this page would be called the home page or the dashboard page which fetches the data from the service(s) using a leaderboard component. But ideally the leaderboard component is dumb and does not require a dependency to a service, because this always makes testing or showing it in the storybook harder.

If we were to add for example some user actions for filtering the timeframe etc. the leaderboard component would have a callback/output/binding to do that but does not smartly also set the filter for the service. Does this make sense? 

In the design system `/ui` the components should be rather generic and reusable. Having a more specialized component building the leaderboard using the code from the design system which also has a story also makes sense to me. I would like to show the leaderboard component in the storybook for documentation but we do not necessarily have to show the whole home page / dashboard page smart component in there.

webapp/src/app/components/leaderboard/leaderboard.component.html
FelixTJDietrich: I guess that's a neat thing to have but IMO it would only make sense if it also includes some change indicators, see for example here: https://willtrumpwin.com/ 
image

For now I would just remove it.

webapp/src/app/components/leaderboard/leaderboard.component.html
GODrums: Really like this suggestion! Added this in 7472b20. 
Unfortunately, this meant that I had to switch from `record` to `class` for the `LeaderboardEntry` since the rank is not known on initialization.

webapp/src/app/components/leaderboard/leaderboard.component.html
GODrums: I think I understand where you are going with that idea now. I personally didn't think about that separation right away because we don't really plan on creating multiple leaderboards rather than one for all repos.

With your explanation I do see the point for it, although I feel like it creates an (unnecessary?) level of abstraction in this use case. Let's discuss further this on Monday maybe and rework in a follow-up PR?

webapp/src/app/components/leaderboard/leaderboard.component.ts
GODrums: Done in 7472b20.

webapp/src/app/components/leaderboard/leaderboard.component.html
FelixTJDietrich: Let's do it in a follow-up. In my mind the leaderboard is just one component on the page. We will add more gamification elements on the same page but you are also right that we might create a leaderboard route per repo in the near future 🤔

webapp/src/app/components/leaderboard/leaderboard.component.html
FelixTJDietrich: This is fine for now, in the future we might directly fetch the entries from the database like some kind of time series. Then we could directly fetch the entries from the db where we can use the rank we get through sorting by score

FelixTJDietrich (approved):
LGTM

PR #84, opened 2024-09-14
Remove example code

FelixTJDietrich:
### Motivation

Code is now unused, we have now existing code which we can

### Description

Remove counter and hello components and the server code associated with it.


GODrums:
LGTM! Would love it, if you could leave it in until the Leaderboard PR is merged though as I'm using the Hello to test my server right now sometimes 😄

FelixTJDietrich:
It is stacked on your PR, of course I will merge it afterwards :)

PR #100, opened 2024-09-20
Leaderboard: Custom Sliding Time Window

GODrums:
### Motivation

We want to be able to set custom timeframes for the leaderboard. This allows us to create sliding time windows with custom dates for the client.

### Description

This PR adds support for custom sliding time windows in both the server- and client-side:
- `before` and `after` timestamps, both usable standalone or combined
- Server-side via `@RequestParam`
- Client-side via URL query params

Example URL: http://localhost:4200/?after=2024-09-18T10:15:30%2001:00&before=2024-09-20T10:15:30%2001:00
-> sets `after` to: 2024-09-19 10:15:30 UTC +1
-> sets `before` to: 2024-09-20 10:15:30 UTC +1

In the future, this should be combined with UI elements (e.g. button, range-slider) to allow the visitor of the website to filter the timeframe they would like to see.

#### General

#### Client (if applicable)

#### Server (if applicable)


webapp/src/app/home/home.component.ts
```diff
@@ -13,9 +14,21 @@ import { lastValueFrom } from 'rxjs';
 export class HomeComponent {
   leaderboardService = inject(LeaderboardService);
 
+  // timeframe for leaderboard
+  // example: 2024-09-19T10:15:30+01:00
```
FelixTJDietrich, line 18: Do we need the timestamp? Wouldn't date be enough?

server/application-server/src/main/java/de/tum/in/www1/hephaestus/leaderboard/LeaderboardService.java
```diff
@@ -99,6 +103,10 @@ public List<LeaderboardEntry> createLeaderboard() {
         return leaderboard;
     }
 
+    private boolean isInTimeframe(OffsetDateTime date, Optional<OffsetDateTime> before, OffsetDateTime after) {
```
FelixTJDietrich, line 106: I think the parameter order can be more understandable by having `after` before `before`

server/application-server/src/main/java/de/tum/in/www1/hephaestus/leaderboard/LeaderboardService.java
```diff
@@ -55,8 +59,8 @@ public List<LeaderboardEntry> createLeaderboard() {
             Set<PullRequestReviewDTO> commentSet = new HashSet<>();
 
             user.getReviews().stream()
```
FelixTJDietrich, line 61: I think `user.getReviews` could already support querying by timeframe.

webapp/src/app/home/home.component.ts
```diff
@@ -13,9 +14,21 @@ import { lastValueFrom } from 'rxjs';
...
+      this.after = params.get('after')?.replace(' ', '+') ?? undefined;
+      this.before = params.get('before')?.replace(' ', '+') ?? undefined;
+    });
+  }
+
   query = injectQuery(() => ({
     queryKey: ['leaderboard'],
-    queryFn: async () => lastValueFrom(this.leaderboardService.getLeaderboard()),
+    queryFn: async () => lastValueFrom(this.leaderboardService.getLeaderboard(this.before, this.after)),
     gcTime: Infinity
```
FelixTJDietrich, line 32: Just remove this, I like that refetches the data by default after 5min. This keeps it somewhat up-to-date :) Also when you revisit the page after a day or so

server/application-server/src/main/java/de/tum/in/www1/hephaestus/leaderboard/LeaderboardService.java
```diff
@@ -36,14 +37,17 @@ public LeaderboardService(UserService userService) {
...
+    public List<LeaderboardEntry> createLeaderboard(Optional<OffsetDateTime> before, Optional<OffsetDateTime> after) {
         logger.info("Creating leaderboard dataset");
 
         List<User> users = userService.getAllUsers();
-        logger.info("Leaderboard has " + users.size() + " users");
 
-        OffsetDateTime cutOffTime = new Date(System.currentTimeMillis() - 1000 * 60 * 60 * 24 * timeframe)
-                .toInstant().atOffset(ZoneOffset.UTC);
+        OffsetDateTime afterCutOff = after.isPresent() ? after.get()
+                : new Date(System.currentTimeMillis() - 1000 * 60 * 60 * 24 * timeframe).toInstant()
```
FelixTJDietrich, line 46: Maybe we should round to the beginning of the day?

FelixTJDietrich (changes requested):

server/application-server/src/main/resources/application.yml
```diff
@@ -33,7 +33,7 @@ springdoc:
 monitoring:
     runOnStartup: true
     # Fetching timeframe in days
-    timeframe: 7
+    timeframe: 1
```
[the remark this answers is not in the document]
GODrums: Yes, this was not meant to be committed and only used for quick testing (waiting 4 minutes for the 7 day fetch can be pretty annoying).

webapp/src/app/home/home.component.ts
GODrums: I think you might be confusing a few elements here.
`gcTime` only determines how often the request is garbage-collected (so for example while the site is inactive in the background).
Even with `gcTime: Infinity`, there are many instances where the query is re-fetched, e.g. window re-focus, site focus change, site refresh, change in query parameters, navigation, etc., so are we sure, that we want to have an additional garbage collection for the site while in the background?

webapp/src/app/home/home.component.ts
GODrums: Yes, of course. I just used a full example as demonstration. Every string, that is parsable as OffsetDateTime in Java, is valid here.

webapp/src/app/home/home.component.ts
FelixTJDietrich: I see. But, I mean this basically says you don't want to delete the fetched data after 5 min if we no longer use the query, i.e. the component with `injectQuery` is unmounted, so it is cached forever. 

I think you can just remove it 🤔

server/application-server/src/main/java/de/tum/in/www1/hephaestus/leaderboard/LeaderboardService.java
FelixTJDietrich: Or what happens with `2024-09-10`? I guess it will use midnight, then it might be important for the `after` case. If you for example use `2024-09-15` as `before` I think it should get all timestamps before the end of day. So actually we would need to round up?

webapp/src/app/home/home.component.ts
```diff
@@ -16,19 +17,14 @@ export class HomeComponent {
...
-  }
+  private readonly route = inject(ActivatedRoute);
+  private queryParams = toSignal(this.route.queryParamMap, { requireSync: true });
+  protected after = computed(() => this.queryParams().get('after')?.replace(' ', '+') ?? undefined);
+  protected before = computed(() => this.queryParams().get('before')?.replace(' ', '+') ?? undefined);
 
   query = injectQuery(() => ({
-    queryKey: ['leaderboard'],
-    queryFn: async () => lastValueFrom(this.leaderboardService.getLeaderboard(this.before, this.after)),
+    queryKey: ['leaderboard', { after: this.after, before: this.before }],
```
FelixTJDietrich, line 26: Don't you have  to call the signal? I.e. `after()` and `before()`

FelixTJDietrich (changes requested):

server/application-server/src/main/java/de/tum/in/www1/hephaestus/leaderboard/LeaderboardService.java
GODrums: So if I understand your idea correctly, you would like to do the filtering in the SQL-Query already?

That's a great idea for performance improvments, I will take a look at it.

server/application-server/src/main/java/de/tum/in/www1/hephaestus/leaderboard/LeaderboardService.java
GODrums: If no timestamp is provided, it should default to midnight, yes.

While I personally find different default timestamps for `before` and `after` confusing, I do see why this would be expected behavior for users. Will change it 👍

server/application-server/src/main/java/de/tum/in/www1/hephaestus/leaderboard/LeaderboardService.java
GODrums: I tried out a Query like this:
```sql
SELECT u
FROM User u
JOIN FETCH u.pullRequests 
JOIN FETCH u.issueComments
JOIN FETCH u.reviewComments
JOIN FETCH u.reviews re
WHERE re.createdAt between :after and :before
OR re.updatedAt between :after and :before
```
but it doesn't seem to return any results, most likely I'm missing something regarding the format conversion.

server/application-server/src/main/java/de/tum/in/www1/hephaestus/leaderboard/LeaderboardService.java
GODrums: Had a look through Java best practices and found this sheet: https://i.sstatic.net/WyAl2.png

Equivalently to your idea, before and after are now only dates (`LocalDate` - without timestamp) with with an interpretation as the start and end of the day on the server respectively. Also looks much cleaner in the URL (example: http://localhost:4200/?after=2024-09-21&before=2024-09-23).

Implemented this change in b76ed6d.

server/application-server/src/main/java/de/tum/in/www1/hephaestus/leaderboard/LeaderboardService.java
GODrums: Managed to find the correct SQL query to filter the dates.
With 63c434d, we now filter directly in the DB.

FelixTJDietrich (approved):
Code LGTM, thank you! 😄

PR #105, opened 2024-09-27
GUI for Custom Time Sliding Window in Leaderboard

GODrums:
### Motivation

After the introduction of sliding time windows on the server in #100, we now aim to provide an intuitive GUI on the client to navigate through the past weekly leaderboards.

### Description

This PR introduces a FILTER-card for the leaderboard among smaller QoL improvements:
- When the data is still available yet, we feature a new `Skeleton`-component with an artificial delay (1s) added on top to improve UX and display smooth transitions
- A filter-card features a new `Select`-component for choosing the timeframe to be displayed in the leaderboard. A timeframe is defined as a calendar week (monday-sunday). The options consist of the current as well as the last 4 weeks.
- The filter card is displayed to the left of the leaderboard or above it on small devices.

### Screenshots (if applicable)

Screenshot 2024-09-27 173746
Screenshot 2024-09-27 175816

#### General

- [x] PR title is clear and descriptive
- [x] PR description explains the purpose and changes
- [x] Code follows project coding standards
- [x] Self-review of the code has been done
- [x] Changes have been tested locally
- [x] Screenshots have been attached (if applicable)
- [ ] Documentation has been updated (if applicable)

#### Client (if applicable)

- [x] UI changes look good on all screen sizes and browsers
- [x] No console errors or warnings
- [x] User experience and accessibility have been tested
- [x] Added Storybook stories for new components
- [x] Components follow design system guidelines (if applicable)

PR #92, opened 2024-09-16
Add pull requests component

iam-flo:
### Motivation
Adds a pull request widget component and the necessary data to display pull requests of a user

### Description
- extended Pull request model and api
- add widget for pull request

### Screenshots (if applicable)

Bildschirmfoto 2024-09-16 um 18 06 58

#### General

- [x] PR title is clear and descriptive
- [x] PR description explains the purpose and changes
- [x] Code follows project coding standards
- [x] Self-review of the code has been done
- [x] Changes have been tested locally
- [x] Screenshots have been attached (if applicable)
- [x] Documentation has been updated (if applicable)

#### Client (if applicable)

- [x] UI changes look good on all screen sizes and browsers
- [x] No console errors or warnings
- [x] User experience and accessibility have been tested
- [x] Added Storybook stories for new components
- [ ] Components follow design system guidelines (if applicable)

#### Server (if applicable)

- [x] Code is performant and follows best practices
- [x] No security vulnerabilities introduced
- [x] Proper error handling has been implemented
- [x] Added tests for new functionality
- [x] Changes have been tested in different environments (if applicable)

PR #108, opened 2024-10-02
Fix: Initial Leaderboard filter params

### LINKED ISSUE #107: Timeframe filter confusing when initially opening the leaderboard
**Describe the bug**
When initially opening the leaderboard, it still looks like the time frame selector is applied. However, when switching to another time frame and back again to the original one, different data is shown (and now the browser URL shows the from/to URL parameters).
I therefore suspect the filter is not applied when loading the page initially.

**Expected behavior**
The UI should make clear that the filter is not yet applied (e.g. by having an empty dropdown option) since the current behaviour is confusing and is not clear if the filter is already applied on first page load or not.

**Screenshots**
URL is `https://hephaestus.ase.cit.tum.de/` without parameters, but filter looks like is is active:
image
Data in the table does not seem to be filtered, however.

**Additional context**

GODrums:
### Motivation

Closes #107.

### Description

We now use the current week as default whenever there are no query params. This should have been part of the original PR #105, but I must have missed it.
It also fixes the Dayjs package version, which got merged with a `^` in #92.

#### General

#### Client (if applicable)

- [ ] UI changes look good on all screen sizes and browsers
- [ ] User experience and accessibility have been tested

### DIFF
--- a/webapp/package-lock.json
+++ b/webapp/package-lock.json
@@ -24,7 +24,7 @@
         "autoprefixer": "10.4.20",
         "class-variance-authority": "0.7.0",
         "clsx": "2.1.1",
-        "dayjs": "^1.11.13",
+        "dayjs": "1.11.13",
         "lucide-angular": "0.429.0",
         "postcss": "8.4.41",
         "rxjs": "7.8.1",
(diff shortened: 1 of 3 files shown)


FelixTJDietrich (approved):
LGTM

PR #113, opened 2024-10-04
Support multiple Repositories in Leaderboard

GODrums:
### Motivation

Many Artemis developers work on multiple related repositories simultaneously. To be able to count all those contributions, we want to be able to view leaderboards for a pre-determined list of repositories as well as a combined leaderboard for all repos.

### Description

This PR makes it possible to distinguish between repositories for the leaderboard:
- New select-based filter to filter for a specific repository or display all
- Leaderboard endpoint now has optional repository parameter, filtering on SQL-level
- Replaced input propagation for URL params with component-based signals to avoid long propagation chains. Unfortunately, this leads to its Storybook story currently having broken controls.
- New error state whenever the leaderboard has no entries
- Spring-Env for production has been adjusted to fetch the following list of repositories: Artemis, Athena, Hephaestus, Pyris, Ares2, Aeolus, hades, Apollon, Apollon_standalone

Repository filter on desktop screen:
Screenshot 2024-10-05 003810
Repository filter on mobile:
Screenshot 2024-10-05 004027
Error state (no entries in a leaderboard):
Screenshot 2024-10-05 002258

#### General

#### Client (if applicable)

- [x] Components follow design system guidelines (if applicable)

#### Server (if applicable)


webapp/src/app/home/leaderboard/filter/repository/repository.component.ts
```diff
@@ -0,0 +1,71 @@
+import { Component, computed, effect, signal } from '@angular/core';
+import { FormsModule } from '@angular/forms';
+import { Router, RouterLink } from '@angular/router';
+import { BrnSelectModule } from '@spartan-ng/ui-select-brain';
+import { HlmSelectModule } from '@spartan-ng/ui-select-helm';
+import { HlmLabelModule } from '@spartan-ng/ui-label-helm';
+
+export const repositoryNames = [
+  'ls1intum/Artemis',
```
FelixTJDietrich, line 9: Can you move this into the environment? Or actually, I think it would be better if we get the available repositories from the server 🤔

server/application-server/src/main/resources/application-prod.yml
```diff
@@ -14,7 +14,7 @@ spring:
       password: ${SECURITY_USER_PASSWORD}
 
 monitoring:
-  repositories: ls1intum/Artemis
+  repositories: ls1intum/Artemis, ls1intum/Athena, ls1intum/Hephaestus, ls1intum/Pyris, ls1intum/Ares2, ls1intum/Aeolus, ls1intum/hades, ls1intum/Apollon, ls1intum/Apollon_standalone
```
FelixTJDietrich, line 17: We could also move this into a environment variable now

FelixTJDietrich (approved):
Code looks good, I only have two remarks

webapp/src/app/home/leaderboard/filter/repository/repository.component.ts
GODrums: I thought about this as well, but does it actually make sense to have individual client-sided requests for small UI elements? Do we plan on changing the repository-list often?

I remember we talked about this when implementing the timeframe-filter and I'm personally not convinced it actually make sense. In NextJS this could be solved via Server-Components (SSR), but I don't think there is an equivalent in our current stack.

Maybe we later on create a config-endpoint to get all of these UI content data points in one request (similar to how i18n works)?

server/application-server/src/main/resources/application-prod.yml
GODrums: Isn't the `application`-file our environment equivalent? Or am I confusing concepts here?

webapp/src/app/home/leaderboard/filter/repository/repository.component.ts
GODrums: Just created the `GET /meta`-endpoint according to what we discussed yesterday. For now it only returns the repositories env-var, but we can recycle it for the adjustable configs, which I'll work on after this.

webapp/src/app/home/leaderboard/filter/repository/repository.component.ts
```diff
@@ -0,0 +1,74 @@
+import { Component, computed, effect, input, signal } from '@angular/core';
+import { FormsModule } from '@angular/forms';
+import { Router, RouterLink } from '@angular/router';
+import { BrnSelectModule } from '@spartan-ng/ui-select-brain';
+import { HlmSelectModule } from '@spartan-ng/ui-select-helm';
+import { HlmLabelModule } from '@spartan-ng/ui-label-helm';
+
+export const repositoryNames = [
```
FelixTJDietrich, line 8: This needs to be removed again

FelixTJDietrich (approved):
Looks good now, thanks for this PR!

PR #120, opened 2024-10-16
Fix: Improve filter layout for mobile

GODrums:
### Motivation

On smaller screens (e.g. mobile) the leaderboard filters are sometimes outside of the viewing area.

### Description

This PR gives the filter component more responsiveness for smaller screens.

Width <580px:
image
Width 580-1280px:
image

#### General

#### Client (if applicable)

### DIFF
--- a/webapp/src/app/home/home.component.html
+++ b/webapp/src/app/home/home.component.html
@@ -1,6 +1,6 @@
 <div class="flex flex-col items-center">
   <div class="">
-    <div class="grid grid-cols-1 xl:grid-cols-4 gap-4 xl:gap-8">
+    <div class="grid grid-cols-1 xl:grid-cols-4 gap-y-4 xl:gap-8">
       <div class="space-y-2 col-span-1">
         <div class="flex flex-col gap-2 mb-4">
           <h1 class="text-3xl font-bold">Artemis Leaderboard</h1>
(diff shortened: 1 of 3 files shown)


FelixTJDietrich (changes requested):
Can you also make the container fit the subviews? There is a lot of whitespace to the side 🤔 
image
image

GODrums:
> Can you also make the container fit the subviews? There is a lot of whitespace to the side 🤔

What testing setup did you use to get this whitespace? I can't seem to reproduce it at all.

FelixTJDietrich:
Ah okay, I was just checking Storybook 🤔 Then it's fine

FelixTJDietrich:
But it also looks kind of like this in your PR screenshot 
image

FelixTJDietrich (approved):

GODrums:
> Ah okay, I was just checking Storybook 🤔 Then it's fine, but ideally it should behave similarly

So the reason why it's not limited in Storybook is because we use the Tailwind `container` to limit the max-width of our whole "page body".



> But it also looks kind of like this in your PR screenshot

The image is exactly the "wrapping" break-point from `flex-wrap`. So if you add a few pixels of width, it will snap back into the same row. Looks like this then:
image

PR #129, opened 2024-11-02
`Developer notifications`: Slack bot with weekly post

GODrums:
### Motivation

One of our goals is to integrate Hephaestus into Artemis's workflows as seemlessly as possible. This includes providing developer notifications about results in the leaderboard. As the primary channel for communication in Artemis, we primarily focus on Slack here.

### Description

This PR sets up the new Slack-Bot `HephaestusBot` for usage within our application-server:
- Automatic testing if the bot works on startup (connection, token + auth)
- CRON-job for a leaderboard message composed with Slack's Block Kit
- Env-vars for: disabling the scheduled message, cron schedule, Slack bot token, Slack signing secret, Channel to send messages in
- Leaderboard weeks now start at Tuesday 9am and last exactly 7 days
- Additionally: Adjusts the timeframe-filter of the leaderboard (see screenshot below)

Example of the scheduled Slack message:
Screenshot 2024-11-07 005837
Screenshot 2024-11-08 232353

#### General

#### Server (if applicable)


server/application-server/src/main/java/de/tum/in/www1/hephaestus/leaderboard/SlackMessageService.java
```diff
@@ -57,26 +63,47 @@ public void sendMessage(String channelId, List<LayoutBlock> blocks, String fallb
...
                 section(section -> section.text(markdownText(
-                        "Here is the *review table* summary for the last week. The table is taken directly from <https://hephaestus.ase.cit.tum.de/|Hephaestus> and now also includes non-review comments & code comments in addition to all variants of reviews for each developer. The score is a weighted heuristic based on the complexities of the reviewed PRs (similar to the old algorithm)."))),
+                        "If you would like to check out more leaderboards or statistics, head over to last week's leaderboard or this month's leaderboard."))),
                 section(section -> section.text(markdownText(
-                        "If you would like to check out more leaderboards or statistics, head over to last week's leaderboard or this month's leaderboard."))));
+                        "The top 3 reviewers of the last week are:"))),
```
FelixTJDietrich, line 95: How about:
`Special thanks to our top 3 reviewers of last week:`

server/application-server/src/main/java/de/tum/in/www1/hephaestus/leaderboard/SlackMessageService.java
```diff
@@ -57,26 +63,47 @@ public void sendMessage(String channelId, List<LayoutBlock> blocks, String fallb
...
+                section(section -> section.text(markdownText(
+                        IntStream.range(0, top3reviewers.size())
+                                .mapToObj(i -> ("• *" + (i + 1) + ". @" + top3reviewers.get(i).user().name() + " *"))
```
FelixTJDietrich, line 98: Do you think this user matching algorithm is enough? Maybe we can be a bit smarter here? For example, my GitHub name is `Felix T.J. Dietrich` but on Slack I'm just `Felix Dietrich`

FelixTJDietrich (changes requested):

server/application-server/src/main/java/de/tum/in/www1/hephaestus/leaderboard/SlackMessageService.java
```diff
@@ -0,0 +1,133 @@
...
+            return;
+        }
+
+        var top3reviewers = getTop3SlackReviewers();
+        logger.info("Top 3 Accounts of the last week: " + top3reviewers);
+
+        logger.info("Sending scheduled message to Slack channel...");
+        List<LayoutBlock> blocks = asBlocks(
+                header(header -> header.text(
+                        plainText(pt -> pt.text(":newspaper: Developer Reviews of the last week :newspaper:")))),
```
FelixTJDietrich, line 86: Remove `Developer` IMO

server/application-server/src/main/java/de/tum/in/www1/hephaestus/leaderboard/SlackMessageService.java
```diff
@@ -0,0 +1,133 @@
...
+                divider(),
+                context(context -> context
+                        .elements(List.of(markdownText("*October 31, 2024* | Hephaestus Announcement")))),
```
FelixTJDietrich, line 89: I think we can also remove the `Hephaestus Announcement`, we already have the bot named Hephaestus

server/application-server/src/main/java/de/tum/in/www1/hephaestus/leaderboard/SlackMessageService.java
```diff
@@ -0,0 +1,133 @@
...
+                section(section -> section.text(markdownText(
+                        "The top 3 reviewers of the last week are:"))),
+                section(section -> section.text(markdownText(
+                        IntStream.range(0, top3reviewers.size())
+                                .mapToObj(i -> ("• *" + (i + 1) + ". @" + top3reviewers.get(i).user().name() + " *"))
+                                .reduce((a, b) -> a + "\n" + b).orElse(""))))
+
+        );
+        try {
+            sendMessage(channelId, blocks, "Developer Reviews of the last week");
```
FelixTJDietrich, line 103: I think we can also remove `Developer` here? How will this be displayed?

server/application-server/src/main/java/de/tum/in/www1/hephaestus/leaderboard/SlackMessageService.java
```diff
@@ -57,26 +63,47 @@ public void sendMessage(String channelId, List<LayoutBlock> blocks, String fallb
         }
     }
 
+    private List<LeaderboardEntryDTO> getTop3SlackReviewers() {
+        LocalDate after = LocalDate.now().minusDays(8);
+        LocalDate before = LocalDate.now().minusDays(1);
```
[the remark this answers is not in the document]
GODrums: Kinda depends on how we schedule the CRON and the timeframe we want.
The default CRON is set to execute at Tuesday 9am. This means that with `minusDays(1)` - `minusDays(8)` we get exactly Monday to Monday.
To be fair, since we include the last date as well, I should change it to Monday-Sunday to be in sync with the leaderboard.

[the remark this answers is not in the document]
FelixTJDietrich: Wouldn't it make sense to post to post? We can also let the leaderboard go from post to post, i.e. Tuesday 9am to Tuesday 9am

server/application-server/src/main/java/de/tum/in/www1/hephaestus/leaderboard/SlackMessageService.java
GODrums: This is basically a fallback text for very small UI elements. I believe this is used for certain kind of notifications, which cannot display the original message.
I just removed `Developer` from it. I will test it again as soon as we the bot gets the `users:read`-permission.

server/application-server/src/main/java/de/tum/in/www1/hephaestus/leaderboard/SlackMessageService.java
GODrums: The reviewed code was still missing the adaptation to when we are able to use the `users:read`-permission, so feel free to take look at the new version now.
Nevertheless, I don't really see another way of comparing the accounts aside from real names since there is no direct connection. We could probably do some kind of similarlity matching for the names to compensate for such differences?

FelixTJDietrich: I was thinking about string edit distance maybe 🤔

server/application-server/src/main/java/de/tum/in/www1/hephaestus/leaderboard/SlackMessageService.java
```diff
@@ -0,0 +1,148 @@
...
+        ChatPostMessageResponse response = slackApp.client().chatPostMessage(request);
+
+        if (!response.isOk()) {
+            logger.error("Failed to send message to Slack channel: " + response.getError());
+        }
+    }
+
+    private List<User> getTop3SlackReviewers() {
+        LocalDate after = LocalDate.now().minusDays(8);
+        LocalDate before = LocalDate.now().minusDays(2);
```
FelixTJDietrich, line 70: I'm still unsure about this 🤔  IMO you need to include everything from post to post, or at least Tuesday to Monday, also I think you are hardcoding stuff again with this minus 2 days for Tuesday, minus 1 seemed more correct. 

We could potentially include the 9:00am timestamp in the leaderboard query so it really goes from Tuesday 9:00am to Tuesday 9:00am. We will have to think about how to select the timeframes in the filter.

server/application-server/src/main/java/de/tum/in/www1/hephaestus/leaderboard/SlackMessageService.java
```diff
@@ -57,26 +63,47 @@ public void sendMessage(String channelId, List<LayoutBlock> blocks, String fallb
...
+        List<LayoutBlock> blocks = asBlocks(
                 header(header -> header.text(
-                        plainText(pt -> pt.text(":newspaper:  Developer Reviews of the last week  :newspaper:")))),
+                        plainText(pt -> pt.text(":newspaper: Developer Reviews of the last week :newspaper:")))),
                 divider(),
                 context(context -> context
-                        .elements(List.of(markdownText("*October 31, 2024*  |  Hephaestus Announcement")))),
+                        .elements(List.of(markdownText("*October 31, 2024* | Hephaestus Announcement")))),
+                section(section -> section.text(markdownText(
+                        "Another *review leaderboard* has concluded. You can check out your placement at <https://hephaestus.ase.cit.tum.de/|hephaestus.ase.cit.tum.de>. The score is a weighted heuristic based on the complexities of the reviewed PRs (similar to the old algorithm)."))),
```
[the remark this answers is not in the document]
FelixTJDietrich: I'm using `APPLICATION_SERVER_URL` for the production setup, see `/Users/felixdietrich/Documents/Hephaestus/webapp/Dockerfile`. 

You can just create an entry in the `application.yaml` defaulting to `localhost:8080` or so if not set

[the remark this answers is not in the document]
FelixTJDietrich: Actually, I lied. I only use it in the client but we can also add this to the environment of the server

server/application-server/src/main/java/de/tum/in/www1/hephaestus/leaderboard/SlackMessageService.java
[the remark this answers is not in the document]
GODrums: That would be possible as well of course. Although we would have to consider that we are currently only not persisting the time anywhere (LocalDate doesn't store it and neither does the state in the URL), so this would require quite a bit of refactoring.
Let's discuss this idea in tomorrow's meeting maybe?

server/application-server/src/main/java/de/tum/in/www1/hephaestus/leaderboard/SlackMessageService.java
GODrums: So as mentioned in the other comment, we could definitely do this. To make processing results easier, I would also prefer to store leaderboard results directly in the DB, so it's easier to build gamification on top of it.

server/application-server/src/main/java/de/tum/in/www1/hephaestus/leaderboard/LeaderboardService.java
```diff
@@ -43,16 +42,12 @@ public LeaderboardService(
...
             Optional<String> repository) {
         logger.info("Creating leaderboard dataset");
-
-        LocalDateTime afterCutOff = after.isPresent() ? after.get().atStartOfDay()
-                : LocalDate.now().minusDays(timeframe).atStartOfDay();
-        Optional<LocalDateTime> beforeCutOff = before.map(date -> date.plusDays(1).atStartOfDay());
-
-        var afterOffset = afterCutOff.atOffset(ZoneOffset.UTC);
-        var beforeOffset = beforeCutOff.map(b -> b.atOffset(ZoneOffset.UTC)).orElse(OffsetDateTime.now());
+        var afterOffset = after.orElse(LocalDate.now().minusDays(timeframe).atStartOfDay().atOffset(ZoneOffset.UTC));
```
FelixTJDietrich, line 48: Timeframe is also just for monitoring, the timeframe to sync. I think something is incorrect here.

IMO we are missing the following config:
 - `hephaestus.leaderboard.weekly.resetDay`, e.g. `2` for Tuesday
 - `hephaestus.leaderboard.weekly.resetTime`, e.g. `9:00` for 9 A.M.

server/application-server/src/main/java/de/tum/in/www1/hephaestus/leaderboard/SlackMessageService.java
```diff
@@ -0,0 +1,165 @@
...
+            return allSlackUsers.stream()
+                .min((a, b) -> Integer.compare(
+                    LevenshteinDistance.getDefaultInstance().apply(
+                        entry.user().name(), a.getName()),
+                    LevenshteinDistance.getDefaultInstance().apply(
+                        entry.user().name(), b.getName()))).orElse(null);
+        }).filter(user -> user != null).toList();
+    }
+
+    @Scheduled(cron = "${slack.cronSchedule}")
```
FelixTJDietrich, line 106: As commented above we could have it like this:
 - `hephaestus.leaderboard.weekly.reset.day`, e.g. `2` for Tuesday
 - `hephaestus.leaderboard.weekly.reset.time`, e.g. `9:00` for 9 A.M.

And then calculate the scheduled interval from that, otherwise we might misconfigure it

server/application-server/src/main/java/de/tum/in/www1/hephaestus/leaderboard/SlackMessageService.java
```diff
@@ -0,0 +1,165 @@
...
+    public void sendScheduledLeaderboard() {
+        if (!runScheduledMessage) {
+            return;
+        }
+
+        // get date in unix format
+        var currentDate = OffsetDateTime.now().toEpochSecond();
```
FelixTJDietrich, line 113: We should use exact times for the top3 and the leaderboard calculation

server/application-server/src/main/java/de/tum/in/www1/hephaestus/leaderboard/SlackMessageService.java
```diff
@@ -0,0 +1,165 @@
...
+                .build();
+
+        ChatPostMessageResponse response = slackApp.client().chatPostMessage(request);
+
+        if (!response.isOk()) {
+            logger.error("Failed to send message to Slack channel: " + response.getError());
+        }
+    }
+
+    private List<User> getTop3SlackReviewers() {
```
FelixTJDietrich, line 72: Missing documentation

server/application-server/src/main/resources/application-prod.yml
```diff
@@ -5,7 +5,9 @@ spring:
     password: ${DATABASE_PASSWORD}
   jpa:
     hibernate:
+      # TODO: Change to none and use liquibase
```
FelixTJDietrich, line 8: Remove this line

server/application-server/src/main/resources/application-prod.yml
```diff
@@ -28,3 +30,9 @@ monitoring:
 
 github:
   authToken: ${GITHUB_AUTH_TOKEN}
+slack:
```
FelixTJDietrich, line 33: I think we should slowly move configuration under `hephaestus`, i.e. `hephaestus.notification.slack`. Or we keep `slack` at the top level and move some parts into `hephaestus.notification`, i.e. `hephaestus.notification.leaderboard.weeky.enabled`

FelixTJDietrich (changes requested):
Generally looks good but I would not hardcode those values on the client side. Also added some improvement comments. But once they are implemented we can merge this PR.

server/application-server/src/main/java/de/tum/in/www1/hephaestus/leaderboard/LeaderboardService.java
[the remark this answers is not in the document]
GODrums: I'd agree with the idea to remove the `Optional`.
Done in 7e5e9eb.

server/application-server/src/main/java/de/tum/in/www1/hephaestus/leaderboard/SlackMessageService.java
```diff
@@ -0,0 +1,165 @@
...
+
+    private static final Logger logger = LoggerFactory.getLogger(SlackMessageService.class);
+
+    @Value("${slack.channelId}")
+    private String channelId;
+
+    @Value("${slack.runScheduledMessage}")
+    private boolean runScheduledMessage;
+
+    @Value("${spring.url:localhost:8080}")
```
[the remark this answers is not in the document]
GODrums: I used the `spring` section, because it contains other details regarding the "setup" as well, such as the name, Keycloak-URL, ... .
Moved it to `hephaestus.host_url` according to your suggestion in df27a34 now.

server/application-server/src/main/resources/application-prod.yml
GODrums: Since we can easily use the Slack-bot for other use cases as well though, moving the whole Slack-section under `notification` wouldn't be fair.
All notification-specific settings are now under `hephaestus.notification` as suggested (ce267df).

server/application-server/src/main/java/de/tum/in/www1/hephaestus/leaderboard/LeaderboardService.java
GODrums: Since we now removed the `Optional`-part, this will be handled by the client.
So this should be obsolete now anyways :)

server/application-server/src/main/java/de/tum/in/www1/hephaestus/leaderboard/SlackMessageService.java
GODrums: Maybe I misunderstood your previous comments, but what would be the advantage over a CRON expression?
We've been using CRONs for other tasks (e.g. Github sync scheduler) for quite some time without issues now.

FelixTJDietrich: I mean, we only support weekly leaderboards and have a reset week date and time, also this configuration has to match. Just to avoid some errors and to restrict the configuration options I would derive it somehow from the weekly notification settings

server/application-server/src/main/java/de/tum/in/www1/hephaestus/leaderboard/SlackMessageService.java
GODrums: What do you mean by "exact times"?

`currentDate` is only used for the date-part of the message:
image

FelixTJDietrich: I guess I'm just worried that this is running while we do a github sync. In that case the main thread is occupied, but this is something we anyways have to fix.

FelixTJDietrich: Also we have the exact times and don't need to use `now` everywhere

server/application-server/src/main/java/de/tum/in/www1/hephaestus/leaderboard/SlackMessageService.java
GODrums: As discussed on Slack, this is now done through a custom `ThreadPoolTaskScheduler`, which dynamically creates a new scheduler with a CRON expression derived from the env-values for `day` and `time`.
Probably way over-the-top, but should be pretty clean :)

server/application-server/src/main/java/de/tum/in/www1/hephaestus/leaderboard/SlackMessageService.java
GODrums: I personally didn't even consider this, but I totally understand your concern.
In my new approach with the `ThreadPoolTaskScheduler` this is now solved as well since we use the exact time.

webapp/src/app/home/leaderboard/filter/timeframe/timeframe.component.html
```diff
@@ -1,11 +1,21 @@
...
+  <hlm-select [placeholder]="placeholder()" [(ngModel)]="value" name="value">
+    <hlm-select-trigger class="w-56">
+      <hlm-select-value />
+    </hlm-select-trigger>
+    <hlm-select-content class="w-56">
+      @for (option of options(); track option.id) {
+        <hlm-option [value]="option.value">{{ option.label }}</hlm-option>
+      }
+    </hlm-select-content>
+  </hlm-select>
```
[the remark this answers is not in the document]
GODrums: How about something like this? Added it in 05e6c67, but we can change it, if that doesn't fit your idea.
image

server/application-server/src/main/java/de/tum/in/www1/hephaestus/leaderboard/SlackMessageService.java
```diff
@@ -0,0 +1,254 @@
...
+                allSlackUsers = slackApp.client().usersList(r -> r).getMembers();
+            } catch (SlackApiException | IOException e) {
+                logger.error("Failed to get Slack users: " + e.getMessage());
+                return new ArrayList<>();
+            }
+
+            return top3
+                .stream()
+                .map(entry -> {
+                    var exactUser = allSlackUsers
```
iam-flo, line 175: maybe extract the mapping to a separate method to simplify the stream sequence.

webapp/src/app/core/modules/openapi/api/leaderboard.service.ts
```diff
@@ -102,10 +102,16 @@ export class LeaderboardService implements LeaderboardServiceInterface {
...
-    public getLeaderboard(after?: string, before?: string, repository?: string, observe?: 'response', reportProgress?: boolean, options?: {httpHeaderAccept?: 'application/json', context?: HttpContext, transferCache?: boolean}): Observable<HttpResponse<Array<LeaderboardEntry>>>;
-    public getLeaderboard(after?: string, before?: string, repository?: string, observe?: 'events', reportProgress?: boolean, options?: {httpHeaderAccept?: 'application/json', context?: HttpContext, transferCache?: boolean}): Observable<HttpEvent<Array<LeaderboardEntry>>>;
-    public getLeaderboard(after?: string, before?: string, repository?: string, observe: any = 'body', reportProgress: boolean = false, options?: {httpHeaderAccept?: 'application/json', context?: HttpContext, transferCache?: boolean}): Observable<any> {
+    public getLeaderboard(after: string, before: string, repository?: string, observe?: 'body', reportProgress?: boolean, options?: {httpHeaderAccept?: 'application/json', context?: HttpContext, transferCache?: boolean}): Observable<Array<LeaderboardEntry>>;
```
iam-flo, line 105: please add a line between methods, then its easier to read.

webapp/src/app/home/home.component.ts
```diff
@@ -33,15 +33,33 @@ export class HomeComponent {
 
   private readonly route = inject(ActivatedRoute);
   private queryParams = toSignal(this.route.queryParamMap, { requireSync: true });
-  protected after = computed(() => this.queryParams().get('after') ?? dayjs().isoWeekday(1).format('YYYY-MM-DD'));
-  protected before = computed(() => this.queryParams().get('before') ?? dayjs().format('YYYY-MM-DD'));
+  protected leaderboardSchedule = computed(() => {
```
iam-flo, line 36: maybe also lines between computed properties, so it is clearer where each property ends

webapp/src/app/core/modules/openapi/api/leaderboard.service.ts
FelixTJDietrich: This is generated code.

webapp/src/app/home/home.component.ts
GODrums: Just added a few more ones to make it more readable. 
I usually prefer a content-based grouping to reduce the time it takes to find correlated code.

FelixTJDietrich:
@GODrums, I tried to improve the consistency of the config names in [8c9fb70](https://github.com/ls1intum/Hephaestus/pull/129/commits/8c9fb70831b07f81690e8f69e50ab841bf372ad4)

webapp/src/app/home/leaderboard/filter/timeframe/timeframe.component.html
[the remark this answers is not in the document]
FelixTJDietrich: Fine for now

FelixTJDietrich (approved):
Looks good now, thank you

Issue #159, opened 2024-11-13
Legend for the Leaderboard


GODrums:
**Is your feature request related to a problem? Please describe.**
Currently, it might not be obvious what a certain icon from the leaderboard refers to or how the score is calculated. Additionally, the score calculation should match the old algorithm's results more closely.

**Describe the solution you'd like**
We want to provide a legend for the leaderboard, explaining icon and the score calculation, to make sure all developers understand the details of the leaderboard. The score algorithm further requires adaptations in regards to the newly available data, such as code comments, to more closely match the effort invested into a review.

**Tasks**

- [ ] Adjust Score Calculation
  - [ ] Reduce impact of comments on own PRs
  - [ ] Cap points gained by code comments to foster manual reviews
- [ ] Display points for individual reviews on the profile page
- [ ] Add legend below leaderboard (#165)
  - [ ] Explain leaderboard icons
  - [ ] Summarize score calculation algorithm

**Describe alternatives you've considered**
None

**Additional context**
Requested by Stephan

FelixTJDietrich:
We will adjust the score calculation in a separate PR, I will just close this issue now.

PR #167, opened 2024-11-18
Display Review Score on User Profiles

GODrums:
### Motivation

To provide a better insight into the score calculation, we want to display the score of individual reviews on a user's profile.
Solves task 2 of #159.

### Description

This PR adapts the scoring logic:
- Refactor scoring into a `ScoringService`
- Support for evaluating scores for single `PullRequestReview` or `IssueComment` entities
- Display the score of single reviews on user profiles

### Tasks

- [x] Extract the scoring logic
  - [x] Create new `ScoringService` 
  - [x] Add scoring method for `PullRequestReview`
  - [x] Add scoring  method for `IssueComment`
    - [x] To solve: when no PR is available 
  - [x] Use `ScoringService` in  `LeaderboardService`
- [x] Score element in `ReviewActivityCard`
  - [x] Design with Storybook
  - [x] Add score to DTO
  - [x] Implement logic in Angular
- [x] Calculate scorefor profile data

### Testing Instructions

TODO

image
image

#### General

#### Client (if applicable)

#### Server (if applicable)


server/application-server/src/main/java/de/tum/in/www1/hephaestus/leaderboard/LeaderboardService.java
```diff
@@ -166,90 +161,8 @@ private int calculateTotalScore(List<PullRequestReview> reviews, int numberOfIss
...
-                    .filter(review -> review.getAuthor().getId() != review.getPullRequest().getAuthor().getId())
-                    .map(review -> commentWeight * calculateCodeReviewBonus(review.getComments().size(), complexityScore))
-                    .reduce(0.0, Double::sum);
-
-                double issueCommentScore = commentWeight * numberOfIssueComments;
-
-                double interactionScore = approvalScore + changesRequestedScore + commentScore + issueCommentScore;
-                return 10 * interactionScore * complexityScore / (interactionScore + complexityScore);
-            })
+            .map(pullRequestReviews -> ScoringService.calculateReviewScore(pullRequestReviews, numberOfIssueComments))
```
FelixTJDietrich, line 164: Why are you not dependency injecting the service?

FelixTJDietrich, line 164: Why are you not dependency injecting the service?

server/application-server/src/main/java/de/tum/in/www1/hephaestus/leaderboard/ScoringService.java
```diff
@@ -0,0 +1,129 @@
...
+import de.tum.in.www1.hephaestus.gitprovider.pullrequest.PullRequestRepository;
+import de.tum.in.www1.hephaestus.gitprovider.pullrequestreview.PullRequestReview;
+import java.util.List;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Service;
+
+@Service
+public class ScoringService {
```
FelixTJDietrich, line 15: I think you have to remove some of the `static` keywords from the functions. This service should be dependency injected

webapp/src/app/user/review-activity-card/review-activity-card.component.html
```diff
@@ -11,10 +11,16 @@
...
-        {{ relativeActivityTime() }}
-        in
-        {{ this.repositoryName() }} #{{ this.pullRequest()?.number }}
+      <div class="flex items-center justify-between">
+        <span class="text-sm text-muted-foreground first-letter:uppercase">
+          {{ relativeActivityTime() }}
+          in
+          {{ this.repositoryName() }} #{{ this.pullRequest()?.number }}
+        </span>
+        <div class="flex items-center text-violet-500 dark:text-violet-600">
```
FelixTJDietrich, line 20: Can you add a tooltip? To me it was not immediately obvious that this is the score.

webapp/src/app/user/review-activity-card/review-activity-card.component.html
```diff
@@ -11,10 +11,16 @@
...
+          <hlm-icon size="sm" name="lucideAward" />
```
FelixTJDietrich, line 21: Does it make sense to also include this icon next to `Score` in the leaderboard in the same color? 
Maybe move the `text-violet` into the tailwind config to define a custom color for the points so we keep it consistent if we use it multiple times.

server/application-server/src/main/java/de/tum/in/www1/hephaestus/gitprovider/pullrequestreview/PullRequestReviewInfoDTO.java
```diff
@@ -18,22 +19,24 @@ public record PullRequestReviewInfoDTO(
...
-                pullRequestReview.getHtmlUrl(),
-                pullRequestReview.getSubmittedAt());
+            pullRequestReview.getId(),
+            pullRequestReview.isDismissed(),
+            pullRequestReview.getState(),
+            pullRequestReview.getComments().size(),
+            UserInfoDTO.fromUser(pullRequestReview.getAuthor()),
+            PullRequestBaseInfoDTO.fromPullRequest(pullRequestReview.getPullRequest()),
+            pullRequestReview.getHtmlUrl(),
+            (int) ScoringService.calculateReviewScore(List.of(pullRequestReview)),
```
FelixTJDietrich, line 35: Would it be possible to also directly return the int instead of casting it?

server/application-server/src/main/java/de/tum/in/www1/hephaestus/leaderboard/LeaderboardService.java
```diff
@@ -166,90 +161,8 @@ private int calculateTotalScore(List<PullRequestReview> reviews, int numberOfIss
         double totalScore = reviewsByPullRequestId
             .values()
             .stream()
-            .map(pullRequestReviews -> {
```
FelixTJDietrich, line 169: Can you remove the `[Source]` again from the leaderboard legend?

server/application-server/src/main/java/de/tum/in/www1/hephaestus/leaderboard/ScoringService.java
```diff
@@ -0,0 +1,129 @@
...
+    /**
+     * Calculates the code review bonus for a given number of code comments and complexity score.
+     * The bonus is a value between 0 and 2.
+     * Taken from the original leaderboard implementation script.
+     *
+     * @param codeComments
+     * @param complexityScore
+     * @return bonus
+     */
+    public static double calculateCodeReviewBonus(int codeComments, int complexityScore) {
```
FelixTJDietrich, line 89: This can be private?

server/application-server/src/main/java/de/tum/in/www1/hephaestus/leaderboard/ScoringService.java
```diff
@@ -0,0 +1,129 @@
...
+
+    /**
+     * Calculates the complexity score for a given pull request.
+     * Possible values: 1, 3, 7, 17, 33.
+     * Taken from the original leaderboard implementation script.
+     *
+     * @param pullRequest
+     * @return score
+     */
+    public static int calculateComplexityScore(PullRequest pullRequest) {
```
FelixTJDietrich, line 111: This can also be private?

server/application-server/src/main/java/de/tum/in/www1/hephaestus/gitprovider/pullrequestreview/PullRequestReviewInfoDTO.java
FelixTJDietrich, line 35: I guess that's why some people usually have a conversion service so we don't have to make it static. 🤔 
I don't know if it is problematic

FelixTJDietrich (changes requested):
I have a few minor remarks. Regarding the `static` `ScoringService` idk if this will be problematic. I understand why you did it but maybe this is where we'd need a DTO conversion service so we can dependency inject the service 🤔

server/application-server/src/main/java/de/tum/in/www1/hephaestus/gitprovider/pullrequestreview/PullRequestReviewInfoDTO.java
GODrums: Although it's not super pretty, the `ScoringService` is pretty much stateless, so making it `static` isn't really an issue here.
I do see the general issue though.

So you would prefer creating a Converter over the `static`-approach here?

GODrums: In the `LeaderboardService` we sum up the scores, so returning `double` makes the sum much more accurate

server/application-server/src/main/java/de/tum/in/www1/hephaestus/leaderboard/LeaderboardService.java
GODrums: Done in f0412c8.

webapp/src/app/user/review-activity-card/review-activity-card.component.html
GODrums: Absolutely (commit 403bf25).
I had a look at our current color system (https://primer.style/foundations/primitives/color) and switched to `github-done-foreground`, which looks extremely similar.
Also the icon next to the score looks good:
image

webapp/src/app/user/review-activity-card/review-activity-card.component.html
GODrums: Done in 01b1fa3.

FelixTJDietrich (approved):
Code looks good to me now, thanks for implementing the requested changes so quickly!

It also looks really good, I really like this addition 😄

PR #168, opened 2024-11-19
Intelligence Service MVP Java Spring Boot setup 

milesha:
Contains #169 

### Motivation
We want to perform and store chats with the AI Mentor.

### Description
This PR adds following entities and functionality:
- `Session`. You can:
   - create a new session
   - get a list of all existing sessions for a given user
- `Message`. You can:
   - get a list of all existing messages in a given session
   - send a message into a session and get response from an OpenAI instance 

### Testing Instructions
- Start the `application-server` and `intelligence service`

#### General

#### Server (if applicable)

- [ ] Proper error handling has been implemented


server/application-server/src/main/java/de/tum/in/www1/hephaestus/chat/message/MessageController.java
```diff
@@ -0,0 +1,33 @@
...
+import org.springframework.web.bind.annotation.RequestBody;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RestController;
+
+import java.util.List;
+
+@RestController
+@RequestMapping("/messages")
+public class MessageController {
+    private final MessageService messageService;
```
iam-flo, line 16: @Autowired

iam-flo (changes requested):

server/application-server/src/main/java/de/tum/in/www1/hephaestus/chat/message/Message.java
```diff
@@ -0,0 +1,43 @@
...
+    @NonNull
+    @Column(name = "sent_at")
+    private OffsetDateTime sentAt = OffsetDateTime.now();
+
+    @NonNull
+    @Enumerated(EnumType.STRING)
+    private MessageSender sender;
+
+    @NonNull
+    @Column(name = "content")
```
GODrums, line 32: Same goes here, if you don't want to specify e.g. the String length or SQL type, you can omit `@Column`.

server/application-server/src/main/java/de/tum/in/www1/hephaestus/chat/message/MessageService.java
```diff
@@ -0,0 +1,84 @@
...
+
+    public MessageService() {
+        ApiClient apiClient = new ApiClient();
+        apiClient.setBasePath("http://127.0.0.1:8000");
+        this.sessionApiClient = new DefaultApi(apiClient);
+    }
+
+    public List<MessageDTO> getMessagesBySessionId(Long sessionId) {
+        return messageRepository.findBySessionId(sessionId).stream()
+                .map(message -> new MessageDTO(message.getId(), message.getSentAt(), message.getSender(),
```
GODrums, line 32: Use `MessageDTO.fromMessage(message)` so the conversion-logic stays within the DTO.

server/application-server/src/main/java/de/tum/in/www1/hephaestus/chat/message/MessageService.java
```diff
@@ -0,0 +1,84 @@
...
+    public MessageDTO sendMessage(String content, Long sessionId) {
+        Optional<Session> session = sessionRepository.findById(sessionId);
+        if (session.isEmpty() || content == null) {
+            return null;
+        }
+
+        Message userMessage = new Message();
+        userMessage.setSender(MessageSender.USER);
+        userMessage.setContent(content);
+        userMessage.setSession(session.get());
```
GODrums, line 46: Maybe just create a new variable for the unpacked session before accessing it via the `Optional` that often

server/application-server/src/main/java/de/tum/in/www1/hephaestus/chat/session/Session.java
```diff
@@ -0,0 +1,38 @@
...
+import org.springframework.lang.NonNull;
+import jakarta.persistence.*;
+import de.tum.in.www1.hephaestus.chat.message.Message;
+import de.tum.in.www1.hephaestus.gitprovider.user.User;
+
+@Entity
+@Table(name = "session")
+@Getter
+@Setter
+@ToString(callSuper = true)
```
GODrums, line 19: Same here, just omit the `callSuper`

server/application-server/src/main/java/de/tum/in/www1/hephaestus/chat/session/Session.java
```diff
@@ -0,0 +1,38 @@
...
+@NoArgsConstructor
+public class Session {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
+
+    @OrderColumn(name = "message_order")
```
GODrums, line 26: I'm probably missing something here, but `Message` doesn't have such a column?
Probably ordering after `sentAt` makes the most sense here?

server/application-server/src/main/java/de/tum/in/www1/hephaestus/chat/session/Session.java
```diff
@@ -0,0 +1,38 @@
...
+    @OneToMany(mappedBy = "session")
+    private List<Message> messages = new ArrayList<>();
+
+    @NonNull
+    @Column(name = "created_at")
```
GODrums, line 31: Just like in `Message`, you can omit `@Column` and make use of the automatic the naming.

server/application-server/src/main/java/de/tum/in/www1/hephaestus/chat/session/SessionController.java
```diff
@@ -0,0 +1,35 @@
...
+import org.springframework.web.bind.annotation.PostMapping;
+import org.springframework.web.bind.annotation.RequestBody;
+import org.springframework.web.bind.annotation.GetMapping;
+
+import java.util.List;
+
+import org.springframework.http.ResponseEntity;
+
+@RestController
+@RequestMapping("/sessions")
```
GODrums, line 15: Our usual convention is to use the singular in endpoint naming. So `/session`.

Same goes for the `MessageController`.

server/application-server/src/main/java/de/tum/in/www1/hephaestus/chat/session/SessionController.java
```diff
@@ -0,0 +1,35 @@
...
+public class SessionController {
+
+    private final SessionService sessionService;
```
GODrums, line 18: Use `@Autowired` instead and omit the constructor.

server/application-server/src/main/java/de/tum/in/www1/hephaestus/chat/session/SessionRepository.java
```diff
@@ -0,0 +1,14 @@
...
+import org.springframework.data.repository.query.Param;
+import org.springframework.stereotype.Repository;
+
+import java.util.Optional;
+import java.util.List;
+
+@Repository
+public interface SessionRepository extends JpaRepository<Session, Long> {
+
+    Optional<List<Session>> findByUserLogin(@Param("login") String login);
```
GODrums, line 13: If you don't specify `@Query`, you shouldn't need `@Param` either.

Also lists don't need to be optional as you would just get an empty list in case the database returns an empty result set.

server/intelligence-service/app/model.py
```diff
@@ -60,7 +61,7 @@ class State(TypedDict):
     token_counter=model,
     include_system=True,
     allow_partial=False,
-    start_on="human",
+    start_on="human",  # TODO: change to "system"
```
GODrums, line 64: Is this TODO meant for this PR? Otherwise maybe create and link a follow-up issue, so that TODOs don't stay in here forever.

GODrums (changes requested):
I added a few comments regarding our usual conventions in Java otherwise this looks nice already :)

iam-flo (dismissed):

server/intelligence-service/app/main.py
```diff
@@ -10,18 +9,15 @@
...
 
-# Global dictionary to store conversation states
-conversations: Dict[str, dict] = {}
-
 
 class ChatRequest(BaseModel):
-    message: str
-    thread_id: Optional[str] = None
+    session_id: str
+    message_content: str
```
FelixTJDietrich, line 15: This has to be a list of messages am I right?

server/intelligence-service/app/main.py
```diff
@@ -30,23 +26,11 @@ class ChatResponse(BaseModel):
...
-        user_input = request.message
-        result = chat_function(thread_id, user_input, state)
-        state = result["state"]
-        response_message = result["response"]["messages"][-1].content
-        conversations[thread_id] = state
-        return ChatResponse(response=response_message)
+    state = {"messages": []}
+
+    result = send_message(
+        thread_id=request.session_id, input_message=request.message_content, state=state
```
FelixTJDietrich, line 32: `thread_id` is now `session_id` am I right?

server/intelligence-service/app/main.py
```diff
@@ -10,18 +9,15 @@
...
 
 
 class ChatResponse(BaseModel):
-    response: str
-    thread_id: Optional[str] = None
+    session_id: str
+    message_content: str
 
 
 @app.post(
```
FelixTJDietrich, line 23: Just move this endpoint to a separate file and let the send message code live in the body. I think this makes it a bit more complicated than it needs to be. We don't have to do controller + service for the intelligence service IMO just keep the code together for now

server/intelligence-service/app/model.py
```diff
@@ -78,29 +78,13 @@ def call_model(state: State):
 app = workflow.compile(checkpointer=memory)
```
FelixTJDietrich, line 78: Can we rename this? Mentally, for me `app` is reserved to `FastAPI`

server/intelligence-service/app/model.py
```diff
@@ -60,14 +60,14 @@ class State(TypedDict):
     token_counter=model,
     include_system=True,
     allow_partial=False,
-    start_on="human",
+    start_on="system",
 )
 
 
 def call_model(state: State):
     chain = mentor_prompt | model
```
FelixTJDietrich, line 68: I thought you wanted to use LangGraph? Why are you using a chain?

server/intelligence-service/app/model.py
FelixTJDietrich: LangGraph tutorial uses `graph = graph_builder.compile()`

FelixTJDietrich, line 78: I would also remove `checkpointer=memory` for now. If we use `MemorySaver` we would also need `PostgresSaver` for a production environment as I understand and we are not there yet. What is it doing as of now?

server/intelligence-service/app/model.py
```diff
@@ -78,29 +78,13 @@ def call_model(state: State):
...
-
-
-def chat(thread_id: str, input_message: str, state: State):
+def send_message(thread_id: str, input_message: str, state: State):
     config = {"configurable": {"thread_id": thread_id}}
     # append the new human message to the conversation
     state["messages"] += [HumanMessage(input_message)]
-    state["step"] += 1
 
     output = app.invoke(
```
FelixTJDietrich, line 86: I think you would use a initial LLM call to choose between the step given the context and then use `conditional_edges` to select between the tasks in a second step. Each task node would then have a different prompt.

FelixTJDietrich (changes requested):

server/application-server/src/main/java/de/tum/in/www1/hephaestus/chat/message/Message.java
```diff
@@ -0,0 +1,43 @@
...
+    private Long id;
+
+    @NonNull
+    private OffsetDateTime sentAt = OffsetDateTime.now();
+
+    @NonNull
+    @Enumerated(EnumType.STRING)
+    private MessageSender sender;
+
+    @Lob
```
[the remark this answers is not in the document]
milesha: No, I have set it back to `@Column(length = Length.LONG16)` in the latests PR after some extra testing

server/intelligence-service/app/main.py
milesha: Yes, I would change the parameter name in the model for the consistency as well

FelixTJDietrich (approved):
I think it should be good now :)

PR #218, opened 2025-01-04
Add own rank to leaderboard overview

GODrums:
### Motivation

We want to make it as easy as possible for users to find themselves in the leaderboard.

### Description

This PR adds a new entry on top of the leaderboard representing the loggedIn-user. If the user does not appear in the leaderboard yet, it provides a default replacement entry with a rank of #entries + 1.

### Testing Instructions

- run `application-server` and `webapp`
- login with Github and go to the homepage
- check out different leaderboards to see if your entry is on top of the leaderboard and displays correct data

image

#### General

#### Client (if applicable)

## Summary by CodeRabbit

- **New Features**
  - Introduced a new leaderboard overview component for improved display of leaderboard entries.
  - Added dynamic scrolling functionality to navigate to specific ranks within the leaderboard.
  - Enhanced the UI with a league info modal for additional league information.
  
- **Bug Fixes**
  - Improved rendering logic to ensure leaderboard data is displayed only when available.

- **Chores**
  - Removed the outdated leaderboard league component and its associated Storybook configuration.


webapp/src/app/home/leaderboard/leaderboard.component.ts
```diff
@@ -53,6 +53,37 @@ export class LeaderboardComponent {
   leaderboard = input<LeaderboardEntry[]>();
   isLoading = input<boolean>();
 
+  protected adjustedLeaderboard = computed(() => {
```
milesha, line 56: We either need to come up with more precise naming than 'adjusted' or include clear code comments. Otherwise, the meaning of the functions will remain unclear without thoroughly reading through the codebase. 🤔

GODrums: You are absolutely correct.
I just added a comment to clarify its purpose since I couldn't think of a fitting name.

webapp/src/app/home/leaderboard/leaderboard.component.ts
```diff
@@ -53,6 +53,38 @@ export class LeaderboardComponent {
...
+
+    const userInLeaderboard = entries.find((entry) => entry.user.login.toLowerCase() === currentUser.username);
+
+    return [userInLeaderboard ?? this.defaultSelfEntry(), ...entries];
+  });
+
+  defaultSelfEntry = computed(() => {
+    return {
+      user: {
+        login: this.user()?.username,
```
milesha, line 72: `this.user()?` can be changed to `currentUser?` in the `defaultSelfEntry`, right? As we already have this variable saved

GODrums: Not quite since `currentUser` is only a local variable in the computed-function of the leaderboard. I reworked the `defaultSelfEntry` in 0b90335 now with the user as parameter to realize your idea anyways.

milesha (dismissed):
lgtm, thanks for the improvements 👍🏼

FelixTJDietrich:
Looks for me like this, as it is now I think it kind of looks off
image

FelixTJDietrich:
My proposal was to have a dedicated section displaying similar this information separate of the leaderboard but above, similarly to the profile header in the league PR. I don't like the duplication if the row.

FelixTJDietrich:
Actually just integrate it into the league section after it is merged 🤔

GODrums:
> My proposal was to have a dedicated section displaying similar this information separate of the leaderboard but above, similarly to the profile header in the league PR. I don't like the duplication if the row.

I tried doing that first, but it's super hard to get the alignments correctly when having a table row without any column headers. But I do see why duplicating the row looks bad for people at the top of the scoreboard. 
Maybe we instead want to try a different format in the league section?

FelixTJDietrich:
I mean thin one:
image

Why not integrate it there? It does not need to be the exact same as the table row, can be a slightly different layout. You still have your highlighted spot in the leaderboard once you scroll down 🤔

FelixTJDietrich (changes requested):
@GODrums how do we proceed with this PR? Would it be possible just to integrate it left of the current league process? 

There is anyways so much whitespace on the right. I don't want to change the order in the leaderboard. 
image

GODrums:
@FelixTJDietrich Understood! I tried to apply your suggestion and removed the original changes.

I added a small grid to display relevant leaderboard data about oneself: own score, current leaderboard rank and activity data.
Do let me know if this fits your idea better:
image

FelixTJDietrich (approved):
I changed it to be like this now:
image

Score and review activity are not so important there, I just think that the rank is important next to the avatar. You can click on it and then it scrolls you to the row.

Issue #123, opened 2024-10-20
Integrate Spotless for Java


FelixTJDietrich:
**Is your feature request related to a problem? Please describe.**
We need a consistent formatter for Java similar to prettier in JS

**Describe the solution you'd like**
I came across [Spotless](https://github.com/diffplug/spotless) used by the github-api library that we are using.
We should integrate it similarly to prettier and have it part of the code quality workflow.

**Describe alternatives you've considered**
To be considered.

GODrums:
Love this, let's do it! 👍 
Alternatively, there is also a Prettier-plugin for Java, if we are rather interesting in keeping the same formatter for all sub-projects in the repo: https://github.com/jhipster/prettier-java

PR #232, opened 2025-01-24
Create detection LLM endpoint

iam-flo:
### Motivation
Fixes #123. Enables detection of bad PR description practices based on predefined rules in the application server.
### Description
Sends PR description and rules to LLM for evaluation.
Configures LLM to analyze descriptions and detect bad practices.


server/application-server/src/main/java/de/tum/in/www1/hephaestus/activity/ActivityService.java
```diff
@@ -46,7 +48,7 @@ public ActivityDTO getActivity(String login) {
             pullRequestBadPracticeRepository.findAssignedByLoginAndOpen(login);
 
         Map<PullRequest, List<PullRequestBadPracticeDTO>> pullRequestBadPracticesMap = openPulLRequestBadPractices
-            .stream()
+            .stream().filter(pullRequestBadPractice -> !pullRequestBadPractice.isResolved())
```
FelixTJDietrich, line 51: This means that we are only showing the unresolved detected bad practices? So the resolved ones disappear from the UI?

server/application-server/src/main/java/de/tum/in/www1/hephaestus/activity/model/PullRequestBadPractice.java
```diff
@@ -1,29 +1,29 @@
...
-import jakarta.persistence.Entity;
-import jakarta.persistence.Id;
-import jakarta.persistence.JoinColumn;
-import jakarta.persistence.ManyToOne;
-import lombok.Getter;
-import lombok.NoArgsConstructor;
-import lombok.Setter;
-import lombok.ToString;
+
+import jakarta.persistence.*;
```
FelixTJDietrich, line 5: Using star imports is not best practice in Java, I think

server/application-server/src/main/java/de/tum/in/www1/hephaestus/activity/badpracticedetector/PullRequestBadPracticeDetector.java
```diff
@@ -17,13 +24,34 @@ public class PullRequestBadPracticeDetector {
...
+        }
+
+        logger.info("Detected {} bad practices for pull request: {}", detectedBadPractices.size(), pullRequest.getId());
+        return detectedBadPractices;
+    }
 
-        List<PullRequestBadPractice> existingBadPractices = pullRequestBadPracticeRepository.findByPullRequestId(
-            pullRequest.getId()
-        );
+    @Transactional
```
FelixTJDietrich, line 47: Why do we need `@Transactional`?

server/application-server/src/main/java/de/tum/in/www1/hephaestus/activity/badpracticedetector/PullRequestBadPracticeDetector.java
```diff
@@ -17,13 +24,34 @@ public class PullRequestBadPracticeDetector {
     @Autowired
     private PullRequestBadPracticeRepository pullRequestBadPracticeRepository;
 
+    @Autowired
+    private BadPracticeDetectorService detectorApi;
+
     public List<PullRequestBadPractice> detectAndSyncBadPractices(PullRequest pullRequest) {
```
FelixTJDietrich, line 30: Can you add more Java doc strings everywhere? It would make it easier to understand what is expected / intended for each method. This is a general comment for all code

server/application-server/src/main/java/de/tum/in/www1/hephaestus/config/BadPracticeDetectorConfig.java
FelixTJDietrich, line 1: This file should not exist, we only need one intelligence service config. Please reuse the existing one from the mentor, you might need to refactor the name

server/application-server/src/main/java/de/tum/in/www1/hephaestus/gitprovider/pullrequest/PullRequestRepository.java
```diff
@@ -20,6 +22,7 @@ SELECT MIN(p.createdAt)
     )
     Optional<OffsetDateTime> firstContributionByAuthorLogin(@Param("authorLogin") String authorLogin);
 
+    @Transactional
```
FelixTJDietrich, line 25: Why is `@Transactional` needed? Can you remove it? I think there should be a way where it is not needed

server/application-server/src/main/java/de/tum/in/www1/hephaestus/config/BadPracticeDetectorConfig.java
FelixTJDietrich: Use `IntelligenceServiceConfig.intelligenceServiceApi`  and delete this file

server/application-server/src/main/resources/db/changelog/1738063112622_changelog.xml
```diff
@@ -1,8 +1,8 @@
 <?xml version="1.1" encoding="UTF-8" standalone="no"?>
 <databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:pro="http://www.liquibase.org/xml/ns/pro" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/pro http://www.liquibase.org/xml/ns/pro/liquibase-pro-latest.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-latest.xsd">
     <changeSet author="iam-flo" id="1738063112622-1">
-        <createTable tableName="pull_request_bad_practice">
-            <column name="id" type="BIGINT">
+        <createTable tableName="pullrequestbadpractice">
```
FelixTJDietrich, line 4: Are you sure about this migration? I think if this builds on the previous migration it will not work as intended. You are never deleting the `pull_request_bad_practice` table and migrating its data. Maybe you also don't need to rename it in the first place. 

Also is the `@NonNull` that you added considered?

server/intelligence-service/app/detector/bad_practice_detector.py
```diff
@@ -0,0 +1,35 @@
...
+    title: str = Field(description="The title of the bad practice.")
+    description: str = Field(description="The description of the bad practice.")
+
+class BadPracticeList(BaseModel):
+    """A list of bad practices detected in a pull request."""
+
+    bad_practices: List[BadPractice] = Field(description="A list of bad practices detected in a pull request.")
+
+
+def detectbadpractices(title, description) -> BadPracticeList:
```
FelixTJDietrich, line 27: I like the snake_case here: `detect_bad_practices`

server/intelligence-service/app/detector/bad_practice_detector.py
```diff
@@ -0,0 +1,35 @@
...
+   prompt_path = Path(__file__).parent / "prompts" / "pullrequest_badpractice_detector.txt"
```
FelixTJDietrich, line 28: I think @milesha created a prompt loader that you can reuse. If you want you can also put the prompt in a `.py` file and just import it as variable. I think that would be even better, it does not need to be in a `.txt` file. That way you can also have multiple prompts that belong together in one file

server/intelligence-service/app/detector/bad_practice_detector.py
```diff
@@ -0,0 +1,35 @@
...
+   with open(prompt_path, "r", encoding="utf-8") as f:
+       prompt_text = f.read()
+   prompt_template = ChatPromptTemplate.from_template(prompt_text)
+   prompt = prompt_template.invoke({"title": title, "description": description})
```
FelixTJDietrich, line 32: Make sure the QA jobs pass

webapp/src/app/user/pull-request-bad-practice-card/pull-request-bad-practice-card.component.html
```diff
@@ -58,7 +58,7 @@
         <brn-collapsible-content>
           @for (badpractice of badPractices(); track badpractice.title) {
             <brn-separator hlmSeparator />
-            <app-bad-practice-card [title]="badpractice.title" [description]="badpractice.description" [resolved]="false"></app-bad-practice-card>
+            <app-bad-practice-card [title]="badpractice.title" [description]="badpractice.description" [resolved]="badpractice.resolved"></app-bad-practice-card>
```
FelixTJDietrich, line 61: Here you can potentially also just pass the DTO type as input so you don't need to do each field separately

FelixTJDietrich (changes requested):
Here are some review comments. Since the migration is now split over two PRs maybe it makes sense to just merge the stacked PRs together after you have implemented the review comments and then we just merge jointly for the production deployment.

server/application-server/src/main/java/de/tum/in/www1/hephaestus/activity/ActivityService.java
iam-flo: Yes, but in the follow up Pr I adjusted the detected logic and there also show resolved bad practices

server/application-server/src/main/resources/db/changelog/1738063112622_changelog.xml
iam-flo: i agree, we should merge  this first and then i will rework this migration

webapp/src/app/user/pull-request-bad-practice-card/pull-request-bad-practice-card.component.html
iam-flo: it is done similarly in the other components. what is the best practice?

PR #238, opened 2025-02-03
Update to poetry 2.0

iam-flo:
Upgrade poetry to version 2.0

To test run 
`poetry self update`
and then
`poetry install`
then test that the application still runs and performs as before.

## Summary by CodeRabbit

- **Chores**
  - Upgraded the dependency management tool to version 2.1.1, ensuring smoother installations and improved performance.
  
- **Documentation**
  - Revised installation instructions to advise users with outdated dependency tool versions to run `poetry self update` before proceeding.
  
- **Refactor**
  - Modernized the project configuration by restructuring dependency definitions and adding explicit Python version compatibility.
  
- **Style**
  - Simplified the configuration settings for Python testing by removing unnecessary environment activation settings.


iam-flo:
need to change workflow as well

GODrums (dismissed):
Tested locally and works.

I used Poetry `v.2.1.1` for testing, which is the most recent version. Maybe we can update this PR to 2.1.1 as well?

.github/workflows/generate-intelligence-service-client.yml
```diff
@@ -43,7 +43,7 @@ jobs:
       - name: Install Poetry
         uses: snok/install-poetry@v1
         with:
-          version: 1.8.5
+          version: 2.0.1
```
FelixTJDietrich, line 46:
```suggestion
          version: 2.1.1
```

.github/workflows/intelligence-service-qa.yml
```diff
@@ -24,7 +24,7 @@ jobs:
       - name: Install Poetry
         uses: snok/install-poetry@v1
         with:
-          version: 1.8.5
+          version: 2.0.1
```
FelixTJDietrich, line 27:
```suggestion
          version: 2.1.1
```

FelixTJDietrich (commented):
I will just update the versions quickly

FelixTJDietrich (approved):
Apparently you can use `poetry run fastapi dev` in the VSCode terminal, TIL.

LGTM

PR #825, opened 2026-03-10
refactor(server): code quality and consistency audit

FelixTJDietrich:
## Summary

- Address 100+ findings from principal-engineer code review
- CRITICAL: Add `@Transactional` to `recalculateUserInternal()`, handle optimistic lock exceptions
- HIGH: Remove dead methods from `AchievementService`, `AchievementRegistry`, `UserAchievementRepository`
- HIGH: Fix registry duplicate bug (deduplicated map values instead of raw records)
- MAJOR: Remove unused controller parameters, fix `@PositiveOrZero`→`@Positive` on `LinearAchievementProgress.target`
- Fix inline styles→Tailwind in all story decorators, CSS lint violations
- Fix diamond league color typo, division-by-zero guards, non-null assertions→nullish coalescing
- Consistent `@/` alias imports, exported prop interfaces, proper type guards
- Biome lint clean: sorted imports, removed unused vars, accessibility fixes

## How to Test

- CI covers this (lint, type-check, build, tests)
- Verify backend compiles: `cd server/application-server && ./mvnw compile`
- Verify frontend compiles: `cd webapp && npx tsc --noEmit`

### DIFF
--- a/webapp/src/components/core/Header.tsx
+++ b/webapp/src/components/core/Header.tsx
@@ -3,6 +3,7 @@ import { Link } from "@tanstack/react-router";
 import { Hammer, LogOut, Settings, User } from "lucide-react";
 import { GitHubSignInButton } from "@/components/auth/GitHubSignInButton";
 import { ModeToggle } from "@/components/core/ModeToggle";
+import { MotionToggle } from "@/components/core/MotionToggle";
 import { SurveyNotificationButton } from "@/components/surveys/survey-notification-button";
 import { Avatar, AvatarFallback, AvatarImage } from "@/components/ui/avatar";
 import { Button } from "@/components/ui/button";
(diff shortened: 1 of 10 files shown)


server/application-server/src/main/java/de/tum/in/www1/hephaestus/achievement/AchievementService.java
```diff
@@ -141,7 +150,16 @@ private AchievementEvaluator resolveEvaluator(AchievementDefinition definition)
...
      */
-    @CacheEvict(value = ACHIEVEMENT_PROGRESS_CACHE, key = "#event.user().get().getId()", condition = "#event.user().isPresent()")
+    @Retryable(
+        retryFor = ObjectOptimisticLockingFailureException.class,
+        maxAttempts = 3,
+        backoff = @Backoff(delay = 100)
+    )
+    @CacheEvict(
+        value = ACHIEVEMENT_PROGRESS_CACHE,
+        key = "#event.user().isPresent() ? #event.user().get().getId() : 0",
```
qpxxl-tum: Double null check is unnecessary since condition ensures user is present. Also mapping to id 0 is not good practise since that id could be another user in the database.

PR #1015, opened 2026-04-14
fix(workspace): keep workspace routes aligned during switching

FelixTJDietrich:
## Description

Fix workspace switching in the webapp by making route params drive workspace-scoped views and replacing manual pathname rewrites with TanStack Router-aware workspace navigation. This keeps the URL, feature gates, redirects, and query keys aligned when moving between multiple workspaces, while falling back to the workspace home for non-portable deep links.

## How to test

1. Open the webapp and switch between at least two workspaces from the sidebar while on the leaderboard, teams, achievements, and admin settings pages.
2. Verify the URL always updates to the selected workspace slug and the page content follows the same workspace.
3. Switch from a mentor thread or user profile page and verify the app lands on a safe route in the target workspace instead of keeping invalid child params.
4. Run `npm run test:webapp`, `npm run build:webapp`, and `npm run check`.

## Summary by CodeRabbit

* **Improvements**
  * Workspace switching now computes destinations more robustly; search params are preserved only when appropriate.
  * Pages and feature flags are explicitly scoped to the current route’s workspace slug, yielding more predictable behavior.
  * Loading, redirect and access flows simplified for more consistent UX.

* **New Features**
  * Added a consolidated workspace-switch planning utility and a navigation hook.

* **Tests**
  * Added tests validating workspace-switching outcomes and targets.


FelixTJDietrich:
Closing — this branch cannot be rebased. `git merge-base origin/main workspace-switching-url-bug` returns nothing: after a history rewrite the branch shares no common ancestor with `main`, so both rebase and merge refuse with *unrelated histories*. The head commit is from 2026-04-15, and the routing it touches has since been reworked by the sync administration (#1350), the admin-gate fix (#1396) and the LLM catalogue (#1400).

**The bug is still real** — `webapp/src/routes/__root.tsx` still rewrites the path by string surgery on switch, so non-portable deep links (mentor threads, user profiles, practice-scoped admin routes) carry across workspaces. It is now tracked as **#1422**, which links this branch as reference for the approach.

**The branch is kept, not deleted.**

PR #1041, opened 2026-04-19
fix(server): aggregate workspace access across linked IdP user rows

FelixTJDietrich:
## Summary

A Keycloak account that federates multiple IdPs (GitHub + GitLab-LRZ) provisions one `user` row per IdP, keyed by `(provider_id, native_id)`. Pre-patch, the resolver looked up the current user by `preferred_username` alone — for a principal whose Keycloak username did not match their GitHub login it silently returned only the GitLab row, so workspaces attached to the GitHub row were invisible. `preferred_username` is operator-settable, which also made the legacy lookup an identity-spoofing surface.

**Production repro:** Keycloak user `ga84xah` linked to `github=felixtjdietrich` (5898705) + `gitlab-lrz=ga84xah` (18024). `findByLogin("ga84xah")` matched only the GitLab row, hiding the `ls1intum` GitHub workspace.

## Changes

**Identity model (extracted to its own service — `UserRepository` no longer reads `SecurityContextHolder`)**
- New `AuthenticatedUserService` — `findAllLinkedUsers()` / `findPrimaryUser()` / `findLinkedUserForProvider()` / `requireCurrentUser()`, driven exclusively by the `github_id` / `gitlab_id` JWT claims emitted by Keycloak IdP protocol mappers. **No `preferred_username` fallback**: a token without identity claims resolves to empty and the caller fails closed.
- Identity lookups in `UserRepository` moved from `ILIKE` to `LOWER(col) = LOWER(:param)` so login/email wildcards cannot widen a match.
- Every caller migrated: `WorkspaceQueryService`, `WorkspaceContextFilter`, `WorkspaceMembershipController`, `AccountController`, `WorkspaceService`, `AuthenticatedGitProviderUserService`, `PracticeFindingService`, `FindingFeedbackService`, `MentorProxyController`.
- Workspace creation picks the linked row whose provider matches `git_provider_mode` so subsequent API calls resolve on the correct IdP identity.

**Multi-IdP aggregation**
- `WorkspaceQueryService.findAccessibleWorkspaces()` unions memberships across all linked rows in a single `IN` query.
- `WorkspaceContextFilter.fetchUserRoles` unions roles via `findAllByWorkspace_IdAndUser_IdIn`. Auto-heal picks the linked row whose provider matches the workspace; if none matches it declines rather than attaching a membership to the wrong IdP. The swallow-all `catch(Exception)` is gone.
- `/members/me` unions memberships and returns the **maximum effective role** (`OWNER > ADMIN > MEMBER`), fixing a prior disagreement where the filter unioned but the controller picked `findFirst`.
- `FindingFeedbackService` scopes access and engagement counts across linked contributor rows.

**JWT claim robustness (`SecurityUtils.readLongClaim`)**
- Unwraps single-element `Collection` claims (Keycloak multivalued-mapper shape).
- Rejects multi-valued / empty / unparseable claims with a sanitized `WARN` log and fails closed.

**Test infrastructure**
- `@WithMockUser` / `@WithUser` / `@WithAdminUser` / `@WithMentorUser` gain `githubId` / `gitlabId` attributes (defaults match the seeded rows in `TestUserConfig`). `TestSecurityConfig`'s mock JWT decoder emits matching claims — so every existing integration test now runs the production claim-driven resolver by default.
- New unit tests: `AuthenticatedUserServiceTest`, `SecurityUtilsJwtClaimsTest`.
- New integration tests: `UserRepositoryProviderLookupIntegrationTest`, `MultiIdpWorkspaceVisibilityIntegrationTest` (reproduces the original bug, locks single-IdP regression, verifies role union).

## Test plan

- [x] `mvn compile` and `mvn test-compile` clean
- [x] New unit tests green (AuthenticatedUserService: 7, SecurityUtilsJwt: 7, FindingFeedbackService: 18)
- [x] No new PMD violations on changed files
- [x] Three principal-engineer review passes applied (final grade A+ with A+ rubric: mandatory gates incl. identity correctness, claim-trust model, architectural layering, case-collation hygiene, test-coverage quality, scope discipline, and malformed-JWT handling all pass)
- [ ] Manual verification on staging/prod: log in as a dual-IdP user and confirm workspaces from both linked rows appear in the switcher and can be entered; single-IdP users unchanged

## Summary by CodeRabbit

* **New Features**
  * Workspace visibility and membership now aggregate across all linked provider identities and pick the highest effective role.

* **Improvements**
  * Safer extraction/parsing of GitHub/GitLab IDs from authentication tokens with warning logs for malformed claims.
  * Workspace creation uses linked-provider identity for owner resolution; client-supplied ownerUserId was removed and ambiguous identities now return conflict.

* **Behavior Change**
  * Claim/merge identity flow is temporarily disabled (HTTP 409) and UI/docs updated.

* **Tests**
  * New unit and integration tests for multi‑IdP scenarios and JWT claim parsing.


FelixTJDietrich:
Addressed the remaining polish batch in  and :\n- hardened  to use claim readers and match GitLab rows by requested server URL\n- stopped logging raw JWT claim values in \n- deleted analytics/tracking data across all linked git-user rows during account deletion\n- resolved mentor proxy user context against the workspace provider-specific linked identity\n- made  fail closed on missing linked identities\n- tightened test helpers and added/updated regression coverage for linked-row role aggregation and feedback aggregation\n\nI also resolved the review nits that are now directly reflected in the branch.\n\nOne review thread remains open intentionally: the  fallback to  when there is no authenticated JWT. That path is still used by direct service-level test/bootstrap callers outside the authenticated controller path, while authenticated HTTP creation already fails closed. I left that behavior unchanged to avoid breaking non-request call sites without a broader refactor.

Issue #1066, opened 2026-05-12
epic: unified Pi runtime for practice detection and interactive mentor


FelixTJDietrich:
## Goal

Consolidate all AI tasks onto the Pi coding agent runtime (`@earendil-works/pi-coding-agent`). PR #1064 moved practice detection onto Pi. This epic extends the same runtime to interactive mentor chat, deletes `server/intelligence-service/`, and leaves the `Task` envelope open for new task kinds (issue review, codebase audits, future non-software domains).

## Shape after the epic

- One Pi image, two runners: `pi-runner.mjs` (one-shot practice) and `pi-mentor-runner.mjs` (long-lived interactive).
- Unified `/workspace/context/` filesystem materialized by `WorkspaceContextBuilder` + `ContentProvider` SPI.
- Sealed `Task` envelope in `task.json` discriminated by `kind`.
- Two sandbox SPIs: `SandboxManager` (existing, one-shot) and `InteractiveSandboxService` (new, long-lived per `(userId, workspaceId)`, stdin/stdout → Spring `SseEmitter`, `AgentSessionRuntime.switchSession` for thread multiplexing).
- Mentor persistence on existing `chat_thread` / `chat_message` / `chat_message_vote` JPA entities, with a new `chat_message.parts JSONB` column replacing the `chat_message_part` table (dual-write, then drop). Resume after container/runner death = `InteractiveSandboxService.attach(...)` brings up a fresh runner; orchestrator calls `runtime.newSession()` for the affected thread, then replays last 20 messages from Postgres via `session.prompt(...)` per message (preceded by one `sessionManager.appendCustomMessageEntry(...)` system note for boundary clarity). Pi's in-container JSONL is container-ephemeral, not a persistence primitive.
- Webapp gets a sibling `useMentorChatPi.ts` hook; the legacy 494-LOC AI-SDK `useMentorChat.ts` is deleted after rollout.

## Key decisions (locked)

| Decision | Choice |
|---|---|
| Transport Java ↔ container | stdin/stdout + `SseEmitter` + virtual threads |
[...]

PR #1086, opened 2026-05-16
feat(mentor): Traefik sticky cookies for replica affinity

### LINKED ISSUE #1077: infra: sticky-cookie Traefik labels for /mentor/chat SSE endpoint
Surfaced during the unified-Pi-runtime planning (#1066). **Blocks #1071 flag rollout in multi-replica environments.**

## Problem

The mentor chat endpoint (`POST /workspaces/{slug}/mentor/chat`, SSE) keeps a `Process` handle for the long-lived Pi container inside the Java replica that spawned it. An SSE reconnect (network blip, browser sleep, etc.) landing on a different replica fails: that replica has no handle, the stream errors.

## Fix

Hephaestus uses Traefik (via Coolify), not nginx-ingress. Add Traefik sticky-cookie labels to the **`application-server`** service (the only service handling `/workspaces/*/mentor/chat`):

```yaml
# docker/compose.app.yaml (production)
services:
  application-server:
    labels:
      traefik.http.services.application-server.loadBalancer.sticky.cookie: "true"
      traefik.http.services.application-server.loadBalancer.sticky.cookie.name: "hephaestus_mentor_session"
      traefik.http.services.application-server.loadBalancer.sticky.cookie.httpOnly: "true"
      traefik.http.services.application-server.loadBalancer.sticky.cookie.secure: "true"
      traefik.http.services.application-server.loadBalancer.sticky.cookie.sameSite: "lax"
```

Apply identically to `docker/preview/compose.app.yaml` for the preview environment.

**Replica identifier exposed for verification.** `application-server` must respond with `X-Hephaestus-Replica: <hostname>` on every response. Add a 10-line `ReplicaIdentityFilter extends OncePerRequestFilter` in this PR — `response.setHeader("X-Hephaestus-Replica", InetAddress.getLocalHost().getHostName())`.

**Cookie lifetime.** Traefik's `sticky.cookie` default is session-lifetime (browser-session). For mentor sessions that should survive a browser refresh, also set `traefik.http.services.application-server.loadBalancer.sticky.cookie.maxAge=3600` (1 h, aligned with our idle TTL × 4). Cookie is per-browser, not per-tab.

**Verification on staging multi-replica.** Staging today may be single-replica (verify by checking Coolify; if so, the test below requires first scaling staging to 2 replicas — a one-time op).

1. Scale staging to 2 replicas (one-line Traefik/Coolify config; coordinate with infra).
2. Open mentor stream, capture `X-Hephaestus-Replica` and the sticky cookie value (`Set-Cookie: hephaestus_mentor_session=…`).
3. Drop the connection (`Ctrl-C` curl or use `kill -STOP` on browser tab).
4. Reconnect with the same cookie. Assert `X-Hephaestus-Replica` returns the same value.
5. Without the cookie (incognito), expect Traefik to round-robin → over N reconnects, different replicas seen.

## Rollback

Remove the labels from `docker/compose.app.yaml` and redeploy. Existing sessions on whichever replica they're pinned to continue until idle eviction; new sessions round-robin. The `Process` handle in the original replica's JVM is still local — without stickiness, reconnects 50% miss and surface as 5xx until the session is restarted. Document this as a known regression of disabling stickiness.

## Out of scope

- Reverse-connection pattern (container as HTTP client to app-server). Future work if multi-region or anycast scenarios arise (out-of-epic; see #1071 FAQ).

## Acceptance criteria

- [ ] Traefik sticky-cookie labels in place on `application-server` in both `docker/compose.app.yaml` and `docker/preview/compose.app.yaml`.
- [ ] `X-Hephaestus-Replica` header present on responses (filter or container-level).
- [ ] Staging multi-replica test (above procedure) passes — reconnect lands on same replica.
- [ ] Without cookie, reconnects round-robin (negative-control test).
- [ ] Documented in `docs/contributor/unified-pi-runtime.mdx`: sticky requirement + how to disable for debugging.
- [ ] Production deployment updated (if multi-replica is live there).

## Dependencies

- Soft blocker for #1071 flag rollout in multi-replica.

FelixTJDietrich:
## Description

Pin user/workspace traffic to the replica holding the in-JVM `PiProcessHandle` so SSE turns don't pay a sandbox cold-start each reconnect. Stamp every response with `X-Hephaestus-Replica` so the affinity is observable from `curl -i` or DevTools.

Fixes #1077

## What changed

| Area | Change |
| --- | --- |
| `docker/compose.app.yaml` | Sticky-cookie + Traefik healthcheck labels on `https-application-server` (the production HTTPS service — HTTP just redirects). Cookie `__Secure-hep_workspace_aff`, `Path=/api/workspaces`, `maxAge=300`, `HttpOnly`, `Secure`, `SameSite=Lax`. |
| `docker/compose.proxy.yaml` | Traefik v3.2 → v3.4 — `sticky.cookie.path` landed in v3.3 ([traefik/traefik#11166](https://github.com/traefik/traefik/pull/11166)); on v3.2 the label is silently dropped. |
| `docker/preview/compose.app.yaml` | Comment-only: previews are single-replica by design and intentionally omit the labels. |
| `server/.../observability/ReplicaIdentityFilter.java` | `OncePerRequestFilter` at `HIGHEST_PRECEDENCE`, emits `X-Hephaestus-Replica` from `$HOSTNAME` (Docker container id) with `InetAddress.getLocalHost()` fallback for non-Docker dev. `@PostConstruct` logs the resolved id once at boot. |
| `server/.../SecurityConfig.java` | CORS `setExposedHeaders(List.of(ReplicaIdentityFilter.HEADER_NAME))` — single source of truth, no drift possible. |
| `docs/contributor/unified-pi-runtime.mdx` | New contributor page covering the affinity story, known limitations, and how to disable. Registered under "AI Development" in the contributor sidebar. |

### Why these specific values

- **`Path=/api/workspaces`**, not `/api/mentor`: the actual mentor URL is `/api/workspaces/{slug}/mentor/chat` (the `WorkspaceScopedController` meta-annotation prefixes `/workspaces/{slug}`). A browser path-prefix match against `/api/mentor` would never trigger; `/api/workspaces` covers the real URL while leaving auth and public endpoints round-robin.
- **`maxAge=300`** matches `hephaestus.mentor.idle-ttl-seconds` (default 300). After the sandbox is reaped the cookie also expires; no pinning to a dead replica.
- **`__Secure-` cookie prefix**: free MITM-rewrite protection given `Secure=true` is already set. `__Host-` is incompatible because of the `Path=/api/workspaces` scope.
- **Traefik backend healthcheck label**: enables Traefik to rewrite the affinity cookie when a replica goes unhealthy. The existing Docker-level healthcheck only triggers container restart.

## Follow-on (folded from #1089/#1090/#1091)

The sticky cookie is one half of the story; the other half is what the JVM is doing with the affinity once Traefik routes it. Items from three rejected standalone PRs are folded in here because they're load-bearing for the same correctness invariant — there is no useful intermediate state where the cookie is in place but the JVM can still race itself onto stripe 0 of 64, replay a stale `agent_end`, pin a carrier thread under load, or silently bury per-user capacity rejections under generic `ERROR`. Each item closes a hole that becomes loud only once the affinity actually fires.

| Item | Why it belongs with replica affinity |
| --- | --- |
| `MentorReplicaAffinityCheck` `HealthIndicator` | Spring Boot can't introspect the upstream LB's routing policy. The sticky cookie is configured one layer up; this readiness probe makes sticky-routing an **operator assertion** (`hephaestus.mentor.expected-replicas` + `hephaestus.mentor.sticky-routing-confirmed`) so a misconfigured multi-replica deploy refuses to receive traffic instead of leaking orphan `IN_FLIGHT` rows behind a 409. Direct complement to `ReplicaIdentityFilter`: one stamps the response, the other refuses to come UP when the contract isn't satisfied. |
| `WorkspaceContextBuilder` mentor striping | `repoKey(ContextRequest)` returned `null` for `MentorChatRequest`, collapsing every concurrent mentor session onto stripe 0 of 64. With affinity now actually pinning chatty users to the same replica, every one of them serialised through the same `ReentrantLock`. Sealed-switch on `ContextRequest` returning `(contributorId * 31) ^ workspaceId` for mentor; 32/64 stripes used over 100 pairs. |
| `AttachedSandbox.subscribe(Cursor)` | `subscribeFromNow` was bolted on as a default method. On a sandbox reused across mentor turns the base `subscribe` replayed the ring and returned the prior turn's `agent_end`, so turn N+1 completed instantly with stale data. Affinity makes sandbox reuse the common case — without this collapse, the second turn on the warm sandbox is what the cookie buys you, and it's broken. |
| Per-thread frame routing via `Predicate<JsonNode>` | The sandbox is shared by `(userId, workspaceId)`; a second chat tab subscribes to the same stream. The post-hoc filter at the consumer (`MentorRunnerClient`) fired the dropped-frame counter for frames the client was never going to deliver. Pushing the predicate into the SPI means rejected frames never enter the per-subscriber queue and never consume drop-counter budget. `MentorFrameFilters.forThread` enforces the broadcast contract (server notifications without a thread destination pass through every filter). |
| `JsonlStdinWriter` `synchronized` → `ReentrantLock` | Mentor turns run on virtual threads. JEP 444's monitor pins the carrier OS thread for the full critical section, including the inner `queue.offer`. Mirrors the same `ReentrantLock` rationale documented on `MentorSseChannel#writeLock`. Becomes a real ceiling once affinity packs more concurrent turns per replica. |
| `FrameRingBuffer` per-user `dropped_total` + 50-user cardinality cap | A misbehaving runner spamming the ring previously ticked one global counter with no per-user attribution. The metric now debounces to ≤1 increment/second per `(userId, sandboxId)` pair and a `MeterFilter` denies the 51st distinct user — bounded cardinality without dropping the global signal. |
| `EvictionReason.MAX_LIFETIME` reaper | Affinity means a chatty power user can stay attached forever and never trip the idle reaper. New `hephaestus.mentor.max-lifetime-minutes` (default 60) is a backstop against accumulated runner state, file-descriptor leaks, and the 165 MB RSS floor × N idle users that motivated the per-user-container sizing in the first place. Wired through the existing `EvictionReason.MAX_LIFETIME` (which was already reserved). |
| `MentorChatMetrics.Outcome.CAPACITY_EXCEEDED` + typed `InteractiveSandboxCapacityExceededException` | Per-user-cap and global-cap rejections used to route through generic `ERROR`. Dashboards conflated "fleet needs scaling" with "runner crashed." Typed SPI exception carries `Scope.PER_USER` / `Scope.GLOBAL`, and `MentorChatService` routes both to a dedicated outcome — alerts split capacity-driven shedding from real failures one tag away. |

## How to test

**Local (CI covers this):**
- `mvn -Dsurefire.includedGroups="unit,architecture" test` — 2685/2685 green (includes the new health check, striping, ring-buffer metric, lifetime reaper, capacity-outcome, predicate-filter, and JsonlStdinWriter concurrency tests).
- `docker compose -f docker/compose.app.yaml -f docker/compose.proxy.yaml config` — resolves all labels cleanly.

**Post-deploy on staging (single-replica is enough for header/CORS smoke):**
1. `curl -i https://<staging>/api/workspaces/<slug>/mentor/chat -d '…'` → assert `X-Hephaestus-Replica: <container-id>` is present.
2. From the webapp in DevTools: `fetch('/api/healthz', { credentials: 'include' }).then(r => r.headers.get('X-Hephaestus-Replica'))` returns the value (proves CORS exposure).
3. `Set-Cookie: __Secure-hep_workspace_aff=…; Path=/api/workspaces; HttpOnly; Secure; SameSite=Lax; Max-Age=300` on responses through Traefik.
4. Hit `/actuator/health` — `mentorReplicaAffinityCheck` reports UP with `expectedReplicas=1, stickyRoutingConfirmed=false` (single-replica safe default).

**Multi-replica (soft blocker, infra-coordinated):**
- Scale staging to 2 replicas (Coolify). Set `hephaestus.mentor.expected-replicas=2` + `hephaestus.mentor.sticky-routing-confirmed=true` on both. `mentorReplicaAffinityCheck` reports UP; leaving `sticky-routing-confirmed=false` flips it to DOWN and Coolify refuses to route — verify both states.
- Capture the cookie + replica on first request; reconnect → same replica. Incognito → round-robin across reconnects.

## Known limitations (documented in `docs/contributor/unified-pi-runtime.mdx`)

- Two browsers, same user → two `(userId, workspaceId)` sandboxes spawned across replicas because `InteractiveSandboxRegistry` is per-JVM.
- Rolling deploys amplify mid-turn disruption for pinned users; drain by refusing new turns on a draining replica and letting in-flight emitters complete (10 min `EMITTER_TIMEOUT_MS`).
- A `MAX_LIFETIME` eviction is unconditional — a user mid-conversation past the 60-minute cap pays one cold start. Tune `hephaestus.mentor.max-lifetime-minutes` if that hurts more than the alternative.

## Out of scope

- Reverse-connection pattern (container as HTTP client → any replica attaches). Tracked under parent epic #1066; not required for #1077.
- `preStop` SIGTERM drain protocol — its own infra issue.
- Workspace-ABI / `pi-mentor-runner.mjs` envelope-mismatch exit code, `PiRunnerProfile` sealed hierarchy, `MentorPiAdapter` `system.md` validation. Those stay scoped to #1088.

## Summary by CodeRabbit

* **New Features**
  * Implemented sticky session cookies for workspace traffic to maintain affinity with the originating application instance
  * Added response header to identify which replica processes each request

* **Chores**
  * Updated Traefik proxy container image to v3.4
  * Added documentation on replica affinity behavior and known limitations
  * Implemented unit tests for replica identification functionality

[Review Change Stack](https://app.coderabbit.ai/change-stack/ls1intum/Hephaestus/pull/1086?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/docker/compose.app.yaml
+++ b/docker/compose.app.yaml
@@ -166,6 +166,18 @@ services:
       - "traefik.http.services.https-application-server.loadbalancer.server.port=8080"
       - "traefik.http.routers.http-application-server.priority=5"
       - "traefik.http.routers.https-application-server.priority=10"
+      # Replica affinity for workspace traffic (mentor lives here) — see docs/contributor/unified-pi-runtime.mdx.
+      # maxAge=300 matches hephaestus.mentor.idle-ttl-seconds; cookie.path requires Traefik >=3.3.
+      - "traefik.http.services.https-application-server.loadbalancer.sticky.cookie=true"
+      - "traefik.http.services.https-application-server.loadbalancer.sticky.cookie.name=__Secure-hep_workspace_aff"
+      - "traefik.http.services.https-application-server.loadbalancer.sticky.cookie.httpOnly=true"
+      - "traefik.http.services.https-application-server.loadbalancer.sticky.cookie.secure=true"
+      - "traefik.http.services.https-application-server.loadbalancer.sticky.cookie.sameSite=lax"
+      - "traefik.http.services.https-application-server.loadbalancer.sticky.cookie.path=/api/workspaces"
+      - "traefik.http.services.https-application-server.loadbalancer.sticky.cookie.maxAge=300"
+      - "traefik.http.services.https-application-server.loadbalancer.healthcheck.path=/actuator/health/liveness"
+      - "traefik.http.services.https-application-server.loadbalancer.healthcheck.interval=10s"
+      - "traefik.http.services.https-application-server.loadbalancer.healthcheck.timeout=3s"
     healthcheck:
       test: "wget -qO- http://localhost:8080/actuator/health || exit 1"
       interval: 5s
--- a/docker/compose.proxy.yaml
+++ b/docker/compose.proxy.yaml
@@ -1,6 +1,6 @@
 services:
   reverse-proxy:
-    image: traefik:v3.2
+    image: traefik:v3.4
     restart: unless-stopped
     networks:
       - shared-network
--- a/docker/preview/compose.app.yaml
+++ b/docker/preview/compose.app.yaml
@@ -178,6 +178,10 @@ services:
   # Application Server - Spring Boot backend
   # Main app domain (set once): https://api.hephaestus.example.com
   # Previews auto-generate: https://{{pr_id}}.api.hephaestus.example.com
+  #
+  # Replica affinity: previews are single-replica by design, so the mentor sticky-cookie
+  # labels from docker/compose.app.yaml are intentionally omitted here — they would be a
+  # no-op. See docs/contributor/unified-pi-runtime.mdx.
   # ---------------------------------------------------------------------------
   application-server:
     build:
--- a/docs/contributor/unified-pi-runtime.mdx
+++ b/docs/contributor/unified-pi-runtime.mdx
@@ -1,0 +1,26 @@
+---
+id: unified-pi-runtime
+sidebar_position: 3
+title: Unified Pi Runtime
+description: Mentor replica affinity and how to disable it for debugging.
+---
+
+## Why mentor traffic is sticky
+
+The mentor SSE endpoint opens a long-lived `docker exec -i` against the user's Pi container (`PiProcessHandle`). The stdin/stdout pipes live in JVM memory; the conversation itself is persisted to Postgres (`chat_thread.session_jsonl`, BYTEA), so any replica can serve any turn — but a replica without the live pipes must rebuild the sandbox first, paying a cold-start cost tracked by `InteractiveSandboxMetrics.attachDuration`.
+
+Traefik pins workspace traffic to the originating replica via a cookie scoped to `/api/workspaces` — narrower than the full `/api` router (auth and public endpoints stay round-robin), broad enough to cover the actual mentor URL `/api/workspaces/{slug}/mentor/chat`. Labels live on the `https-application-server` service in `docker/compose.app.yaml`. Inspect with `curl -i`:
+
+- Cookie: `__Secure-hep_workspace_aff` — `Secure`, `HttpOnly`, `SameSite=Lax`, `maxAge` matches `hephaestus.mentor.idle-ttl-seconds` (300s default) so the cookie expires when the sandbox is reaped.
+- Response header: `X-Hephaestus-Replica: <container-id-prefix>` — emitted by `ReplicaIdentityFilter` from `$HOSTNAME`, CORS-exposed for the webapp.
+
+Labels are HTTPS-only; the HTTP router redirects to HTTPS, so pinning it would just double-issue cookies. The `cookie.path` attribute requires Traefik >= 3.3 (the proxy is pinned to v3.4 in `docker/compose.proxy.yaml`). Previews (`docker/preview/compose.app.yaml`) are single-replica and omit the labels.
+
+## Known limitations
+
+- **Two browsers, same user.** Two browsers can pin to two replicas, each spawning its own `(userId, workspaceId)` sandbox — `InteractiveSandboxRegistry` is per-JVM.
+- **Rolling deploys.** Each pinned user whose replica restarts pays one cold start. The `SseEmitter` timeout is 10 minutes (`MentorChatController.EMITTER_TIMEOUT_MS`); drain by refusing new turns and letting in-flight emitters complete.
+
+## Disabling affinity for debugging
+
+Delete the `traefik.http.services.https-application-server.loadbalancer.sticky.*` labels from `docker/compose.app.yaml` and redeploy. Existing sessions stay pinned until the cookie expires; new sessions round-robin. Expect transient 5xx on reconnects that land on a cold replica.
--- a/docs/sidebars.contributor.ts
+++ b/docs/sidebars.contributor.ts
@@ -25,7 +25,7 @@ const sidebars: SidebarsConfig = {
     {
       type: 'category',
       label: 'AI Development',
-      items: ['ai-agent-workflow', 'ai-code-review'],
+      items: ['ai-agent-workflow', 'ai-code-review', 'unified-pi-runtime'],
     },
   ],
 };
--- a/server/application-server/src/main/java/de/tum/in/www1/hephaestus/SecurityConfig.java
+++ b/server/application-server/src/main/java/de/tum/in/www1/hephaestus/SecurityConfig.java
@@ -2,6 +2,7 @@ package de.tum.in.www1.hephaestus;
 
 import de.tum.in.www1.hephaestus.config.CorsProperties;
 import de.tum.in.www1.hephaestus.feature.FeatureFlag;
+import de.tum.in.www1.hephaestus.observability.ReplicaIdentityFilter;
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
@@ -142,6 +143,7 @@ public class SecurityConfig {
         configuration.setAllowedHeaders(
             List.of("Authorization", "Content-Type", "Accept", "X-Requested-With", "Origin")
         );
+        configuration.setExposedHeaders(List.of(ReplicaIdentityFilter.HEADER_NAME));
         configuration.setAllowCredentials(true);
         configuration.setMaxAge(3600L);
         UrlBasedCorsConfigurationSource source = new UrlBasedCorsConfigurationSource();
--- a/server/application-server/src/main/java/de/tum/in/www1/hephaestus/observability/ReplicaIdentityFilter.java
+++ b/server/application-server/src/main/java/de/tum/in/www1/hephaestus/observability/ReplicaIdentityFilter.java
@@ -1,0 +1,50 @@
+package de.tum.in.www1.hephaestus.observability;
+
+import jakarta.annotation.PostConstruct;
+import jakarta.servlet.FilterChain;
+import jakarta.servlet.ServletException;
+import jakarta.servlet.http.HttpServletRequest;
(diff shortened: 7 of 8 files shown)


FelixTJDietrich:
## Scope reset

I just force-pushed this branch back to its original `2c5a60236` commit, dropping `7090c2e45` ("fix(mentor): correctness + observability adjacent to replica affinity") and `1d2b188d9` (prettier-format follow-up).

Two independent principal-engineer audits agreed that `7090c2e45` was scope creep relative to #1077. PR scope returns to **+114 / -2**. The dropped commit's 1714 lines aren't lost — the SHA is still reachable in the reflog and on `--all` — but they don't belong on this PR.

### Audit verdicts (per piece in `7090c2e45`)

| Piece | Score | Verdict |
|---|---|---|
| `MentorReplicaAffinityCheck` HealthIndicator | 20/50 | REVERT — wired to liveness probe by default (not readiness), guards properties no compose file sets, would kill the pod on a multi-replica deploy unless the operator also sets a flag they have no signal to know about. A 5-line `@PostConstruct` validation would do the job. |
| `InteractiveSandboxMetrics` per-user counters + debounce | 13/50 | REVERT — observability theatre; reinvents Prometheus `rate(1m)` in a `ConcurrentHashMap`; duplicate `interactive_sandbox.evicted` alongside existing `mentor.session.eviction`; 156-line `FrameRingBufferMetricsTest` for a smoothing knob. |
| `MetricsCardinalityConfig` | 22/50 | REVERT — correct Micrometer pattern, exists solely to make the speculative per-user metric safe. |
| `JsonlStdinWriter` `synchronized → ReentrantLock` | 18/50 | REVERT (theatre) — on JDK 21 (which this repo targets) the critical section is just `ArrayBlockingQueue.offer`, a non-blocking op. No measurable pin. The cited JEP 444 risk doesn't apply to non-blocking work inside the monitor; the actual blocking call (`ack.get(timeout)`) was already outside the synchronized block. |
| `WorkspaceContextBuilder` mentor striping hash | 37/50 | SPLIT — **legitimate concurrency bug fix** (mentor sessions all collapsed onto `repoLockStripes[0]` because `repoKey()` returned null). Deserves its own PR where a reviewer can focus on the sealed-switch refactor without sticky-cookie noise. |
| `subscribe(Cursor, Predicate, Consumer)` SPI redesign | 24/50 | SPLIT — fixes a **real** stale-`agent_end` replay race on reused sandboxes, but ships ~400 lines of churn (Cursor enum, Predicate pushdown, four new/expanded test files) for what a `subscribeFromNow()` default-method overload would solve. The Predicate-in-SPI form bleeds JSON-RPC routing conventions (`params.threadId`) into a generic sandbox interface — a strictly worse boundary than keeping `MentorFrameFilters` in the mentor package. |
| `MentorChatMetrics.Outcome.CAPACITY_EXCEEDED` + exception | 31/50 | SPLIT — small, clean, useful for alerting separation. Unrelated to sticky routing. |

### Follow-ups to file separately

Three legitimate items deserve their own PRs (not this one):

1. **`WorkspaceContextBuilder` mentor striping fix** — real concurrency bug, ~80 lines including test.
2. **`subscribeFromNow()` SPI overload** (minimal form, not the Cursor enum) — fixes the stale `agent_end` replay race for reused sandboxes.
3. **`CAPACITY_EXCEEDED` outcome + exception type** — clean alerting separation.

If anyone wants the audit transcripts, I have them and can paste on request.

PR #1090, opened 2026-05-16
fix(server): post-#1081 correctness sweep

FelixTJDietrich:
## Summary

Retrospective PE review identified five correctness bugs where #1081 (server-side Pi mentor over SSE) stretched #1078's runtime kernel and #1080's interactive SPI past their original design. All five fixes ship in this one PR.

| bug | originating PR | stretch direction |
|---|---|---|
| Striping degenerate for mentor (all on stripe 0) | #1078 | mentor never assumed in `repoKey` |
| `subscribeFromNow` bolted on as default method | #1080 | mentor reuses sandbox across turns; replay sees stale `agent_end` |
| `MentorReplicaAffinityCheck` never landed | #1081 | claimed in design, only sticky-cookie enforcement existed |
| `TaskSealedHierarchyTest` deleted | #1078 → #1081 churn | guardrail for permit / `@JsonSubTypes` drift gone |
| `pi-mentor-runner.mjs` not WorkspaceAbi-synced | #1078 → #1081 | second runner script bypassed the cross-language sync test |

## Changes

- **Bug 1** — `WorkspaceContextBuilder.stripeKey` is now a sealed switch: `PracticeReview` stripes on `repository_id` (with job-id fallback), `MentorChat` stripes on `(contributorId, workspaceId)`. New `stripingForMentorRequestsIsNotDegenerate` test asserts ≥ 32/64 distinct stripes over 100 pairs.
- **Bug 2** — Collapsed `AttachedSandbox.subscribe(Consumer)` + `subscribeFromNow(Consumer)` into `subscribe(Cursor, Consumer)`. New top-level enum `Cursor { RING_REPLAY, FROM_NOW }` in `agent/sandbox/spi/`. Every caller now passes an explicit cursor; `MentorRunnerClient` uses `FROM_NOW` to avoid replaying prior turns' `agent_end`. No default method, no shim.
- **Bug 3** — New `MentorReplicaAffinityCheck` `HealthIndicator` reads `${HOSTNAME}`, `hephaestus.mentor.expected-replicas`, `hephaestus.mentor.sticky-routing-confirmed`. Reports DOWN when `replicas > 1 && !confirmed`. Unit test covers all three deploy shapes. Documented in `docs/contributor/agent/mentor-deployment.mdx`.
[...]

PR #1091, opened 2026-05-16
refactor(server): tighten agent runtime + interactive SPI seams

FelixTJDietrich:
## Summary

Stacked on PR #1090 (`post-1081-correctness-sweep`). Tightens four architectural seams stretched by #1078 / #1080 / #1081 while the context is fresh — every change is mechanical, behaviour-preserving, and gated by a test.

### 1. `PiRuntimeFactory.build` is no longer a god method
The 160-line `build()` previously branched on runner kind via filename match (`MENTOR_RUNNER_SCRIPT` constant + `nodeFlagsFor`/`nodeEnvFor`/`shouldRegisterHephaestusProvider` switches). Adding a third runner kind meant editing four switches in the kernel.

- New `PiRunnerProfile` interface in `agent.runtime`; implementations `PracticeRunnerProfile` (in `agent.practice`) and `MentorRunnerProfile` (in `agent.mentor`) own the runner-specific V8 flags + per-process env next to their adapters.
- Kernel now reads three properties off the profile; the filename-match dispatch is gone.
- `shouldRegisterHephaestusProvider(spec)` deliberately stays a kernel-side helper — it's a function of credentials/baseUrl, not runner kind, and both profiles compute it identically.
- The profile interface is **non-sealed**: a `sealed ... permits PracticeRunnerProfile, MentorRunnerProfile` would force `agent.runtime` to import its domain implementations — exactly the dependency the boundary rules forbid. Enforced architecturally instead via a new `PiRunnerProfilePlacement` rule, mirroring the established `ContentProvider` pattern.
- Removed the now-redundant `runnerScript` field from `PiAgentProperties` / `MentorAgentProperties` — it decoupled the script filename from the V8 flags / per-process env it must pair with.

### 2. Per-thread frame routing pushed into the SPI
`MentorRunnerClient` used to filter inbound frames by `boundThreadId` inside `handleEvent`. A future logical-multiplex caller would have had to reinvent the same wheel.

- New `AttachedSandbox.subscribe(Cursor, Predicate<JsonNode>, Consumer)` with server-side filtering: the predicate runs inside `FrameSubscription.offer` before queue enqueue, so rejected frames consume neither queue capacity nor drop-counter budget.
- The two-arg form becomes a default delegating to `frame -> true`.
- `MentorFrameFilters.forThread(threadId)` encapsulates the JSON-RPC-aware filter — broadcast frames (notifications without a `params.threadId`) cross every filter; only frames with a mismatched threadId are dropped.
- SPI Javadoc codifies the broadcast contract.

### 3. `JsonlStdinWriter` no longer uses `synchronized` under virtual threads
Mentor's `send()` runs on virtual threads; the previous `synchronized` blocks risked JEP 444 pinning the carrier OS thread for the full critical section (incl. inner queue ops). Replaced with a `ReentrantLock` — mirrors the precedent set by `MentorSseChannel#writeLock`. Behaviour identical; pin risk gone.

### 4. `PiPlanSpec.extraInputs` path-validation guard
Adapters could previously write to arbitrary workspace paths. Now `extraInputs` keys must be either:
- prefixed by a known dynamic-suffix prefix (`CONTEXT_TARGET_PREFIX`, `SESSIONS_DIR_PREFIX`), or
- listed in `WorkspaceAbi.allowedExtraInputPaths()` (currently `MENTOR_SYSTEM_PROMPT_PATH` from #1090).

Violations fail at construction time (boot / first unit test) with a message naming the offending path. Forces new mount points to be declared on `WorkspaceAbi`.

## Test plan

- [x] `./mvnw test -Dsurefire.includedGroups="unit,architecture"` — **2683 tests** (2556 unit + 127 architecture), all green.
- [x] New tests: `PiRuntimeFactoryTest` parametrised on both profiles, `MentorFrameFiltersTest`, `AttachedSandboxFilterTest`, `JsonlStdinWriterConcurrencyTest`, `PiPlanSpecValidationTest`.
- [x] Verification gates:
  - `git grep -n "synchronized" server/application-server/src/main/java/de/tum/in/www1/hephaestus/agent/sandbox/docker/interactive/JsonlStdinWriter.java` → 0
  - `git grep -n "MENTOR_RUNNER_SCRIPT" server/application-server/src/main/java/de/tum/in/www1/hephaestus/agent/runtime/` → 0
- [x] `AgentRuntimeBoundaryTest` extended with `PiRunnerProfilePlacement` rule — passes.
- [x] No backwards-compat shims; one commit.

## Stacking notes

Base branch is `post-1081-correctness-sweep` (PR #1090). The Cursor enum + `MENTOR_SYSTEM_PROMPT_PATH` constant + correctness fixes from that PR are dependencies, not duplicated here.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/server/application-server/src/main/java/de/tum/in/www1/hephaestus/agent/mentor/MentorAgentProperties.java
+++ b/server/application-server/src/main/java/de/tum/in/www1/hephaestus/agent/mentor/MentorAgentProperties.java
@@ -13,12 +13,15 @@ import org.springframework.validation.annotation.Validated;
 /**
  * When llmProvider + credentialMode + modelName are all set, mentor uses them directly and
  * skips the workspace-scoped AgentConfig table. Omit any one to fall back to AgentConfig.
+ *
+ * <p>The runner script name is owned by {@link MentorRunnerProfile} — operator overrides for it
+ * were never used in practice and risked drifting from the V8 flags / per-process env that the
+ * kernel pairs with the script. Bumping the runner is a code change.
  */
 @Validated
 @ConfigurationProperties(prefix = "hephaestus.mentor.agent")
 public record MentorAgentProperties(
     @DefaultValue("ghcr.io/ls1intum/hephaestus/agent-pi:latest") @NotBlank String image,
-    @DefaultValue("pi-mentor-runner.mjs") @NotBlank String runnerScript,
     @DefaultValue("100000") @Min(1) int maxPromptChars,
     @DefaultValue("") String baseUrl,
     @Nullable LlmProvider llmProvider,
--- a/server/application-server/src/main/java/de/tum/in/www1/hephaestus/agent/mentor/MentorPiAdapter.java
+++ b/server/application-server/src/main/java/de/tum/in/www1/hephaestus/agent/mentor/MentorPiAdapter.java
@@ -35,7 +35,9 @@ public class MentorPiAdapter {
     public static final String ASPECT_INPUT_PREFIX = "context/target/";
 
     /** Workspace-relative directory for restored Pi SDK session JSONL files (matches the runner's {@code SESSIONS_DIR}). */
-    public static final String SESSIONS_DIR_PREFIX = ".sessions/";
+    public static final String SESSIONS_DIR_PREFIX = WorkspaceAbi.SESSIONS_DIR_PREFIX;
+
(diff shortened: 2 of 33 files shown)


FelixTJDietrich:
Folding into #1086 + #1088 — SPI hygiene split between mentor-runtime (#1086) and runtime-kernel (#1088).

PR #1279, opened 2026-05-18
feat(profile): add configurable activity monitor

ValentinGruener:
## Description

Adds a configurable activity monitor to the profile page with improved alignment for review activity and open pull requests. Compared to the previous profile activity section, the monitor now groups the activity lists under one shared heading, aligns both columns consistently, supports repository filtering, and uses the generated OpenAPI client.

## How to test

- Ran `pnpm run format`
- Ran `pnpm run check`
- Start Storybook with `pnpm --filter webapp run storybook`
- Verify:
  - `ActivityMonitorFilled`
  - `ActivityMonitorRepositoryFiltered`

### Default activity monitor

image

### Repository filtered activity monitor

image

## Summary by CodeRabbit

* **New Features**
  * Activity monitor on profiles: aggregated stats, authored PRs, review activity, repository filters, configurable limit, URL-persistent filter state, "View all" actions, configuration popover, and loading skeletons.
  * New profile activity-monitor API and response shape.

* **Style**
  * Updated timeframe picker sizing for improved layout.

* **Tests**
  * Added integration and unit tests covering date-range queries, activity-monitor behavior, and configuration wiring.

* **Chores**
  * New sync feature toggles added (disabled by default).

[Review Change Stack](https://app.coderabbit.ai/change-stack/ls1intum/Hephaestus/pull/1279?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


FelixTJDietrich:
Generally looks good :) I'm doing some other fixes right now to the DX that I noticed while testing locally. Will merge afterwards once I've tested it with my ls1intum data

server/src/main/java/de/tum/cit/aet/hephaestus/profile/UserProfileService.java
```diff
@@ -300,6 +370,111 @@ private List<ProfileReviewActivityDTO> buildReviewActivity(Long userId, Long wor
...
+            changeRequests,
+            comments,
+            codeComments,
+            unknowns,
+            0,
+            authoredPullRequests.size(),
+            0,
+            0,
+            0,
+            0
```
FelixTJDietrich: The 0s for `numberOfOwnReplies`, `numberOfMergedPullRequests`, `numberOfClosedPullRequests`, `numberOfOpenedIssues`, `numberOfClosedIssues` are placeholders — the method doesn't actually compute them. The DTO docs treat those fields as real counts, so the wire shape says "we computed zero" when it's actually "not computed". `ActivityBadges` hides 0 today so nothing visibly breaks, but a different caller would read 0 as truth.

Two ways out: reuse the existing stats computation from `getUserProfile` (then filter by repo on top), or split out a `ProfileMonitorStatsDTO` with only the fields you compute.

server/src/test/java/de/tum/cit/aet/hephaestus/profile/UserProfileServiceTest.java
```diff
@@ -125,9 +125,9 @@ void queriesActivityEventFirstThenHydrates() {
             PullRequestReview review = createReview(400L, user, pr);
 
             when(userRepository.findByLogin(USER_LOGIN)).thenReturn(Optional.of(user));
-            when(profilePullRequestQueryRepository.findAuthoredByLoginAndStates(any(), any(), any())).thenReturn(
-                List.of()
-            );
+            when(
+                profilePullRequestQueryRepository.findAuthoredByLoginAndStates(any(), any(), any(), any(), any())
```
FelixTJDietrich, line 141: No tests reach `getActivityMonitor` — this file only widens existing stubs with extra `any()` args. Could you add cases for:

- `workspaceId` null → empty
- unknown login → empty
- `repositoryIds` filter, hits and misses
- empty `repositoryIds` = all
- `limit` clamp at 0, 101, null
- repository dedup when the same repo appears in both lists
- `totalReviewActivityCount` / `totalAuthoredPullRequestCount` are post-filter, pre-limit

Side note on the widened stubs: matching `any()` for `since` / `until` means a swap of `after` / `before` in the call site would silently pass. Worth tightening at least one existing test to pin those args.

webapp/src/routes/_authenticated/w/$workspaceSlug/user/$username/index.tsx
```diff
@@ -106,10 +146,26 @@ function UserProfile() {
...
 			profileData={profileQuery.data}
+			activityMonitorData={activityMonitorQuery.data}
+			activityMonitorFilters={{
+				repositoryIds: selectedRepositoryIds,
+				limit: monitorLimit,
+			}}
+			onActivityMonitorFiltersChange={handleActivityMonitorFiltersChange}
 			isLoading={
 				(profileQuery.isPending && !profileQuery.data) ||
 				(workspaceQuery.isPending && !workspaceQuery.data)
```
FelixTJDietrich: `activityMonitorQuery.isPending` is missing from this OR. On first paint the monitor section can render with `activityMonitorData=undefined` while `isLoading=false`, which shows empty-state copy instead of skeletons.

Minimal fix: `|| (activityMonitorQuery.isPending && !activityMonitorQuery.data)`. Cleaner: pass a separate `monitorLoading` flag so refetches show a pulse instead of blanking the rows.

webapp/src/components/profile/ProfileContent.tsx
```diff
@@ -49,64 +77,105 @@ export function ProfileContent({
...
 	schedule,
 }: ProfileContentProps) {
-	// Generate skeleton arrays for loading state
 	const skeletonReviews = isLoading ? Array.from({ length: 3 }, (_, i) => ({ id: i })) : [];
-
 	const skeletonPullRequests = isLoading ? Array.from({ length: 2 }, (_, i) => ({ id: i })) : [];
 
+	const monitorReviewActivity = activityMonitorData?.reviewActivity ?? reviewActivity;
+	const monitorPullRequests = activityMonitorData?.authoredPullRequests ?? openPullRequests;
+	const monitorActivityStats = activityMonitorData?.activityStats ?? activityStats;
```
FelixTJDietrich: These `activityMonitorData?.X ?? legacyProp` fallbacks are dead — the route always passes `activityMonitorData`. The legacy props in `ProfilePage` (`reviewActivity`, `openPullRequests`, `activityStats`, `reviewedPullRequests`) can go too.

The real cost of keeping both: when a user filters by repo, the badges read filtered counts from `activityMonitorQuery.activityStats`, while `reviewedPullRequestsForPopover` can fall back to `profileQuery.reviewedPullRequests` (workspace-scoped, unfiltered). Same UI row, two scopes.

If `getUserProfile` took `repositoryIds` + `limit`, the second query and these fallbacks all go away. Any reason for splitting them?

server/openapi.yaml
```diff
@@ -1608,6 +1608,62 @@ paths:
...
+          type: array
+          items:
+            type: integer
+            format: int64
+          uniqueItems: true
+      - in: query
+        name: limit
+        required: false
+        schema:
+          type: integer
```
FelixTJDietrich, line 1654: Add `maxItems: 100` (or 200) to `repositoryIds` and `minimum: 1, maximum: 100` to `limit` so the spec matches `MAX_ACTIVITY_MONITOR_LIMIT`. The server already clamps, but the generated client and external consumers don't see that today.

server/src/main/java/de/tum/cit/aet/hephaestus/profile/UserProfileService.java
```diff
@@ -300,6 +370,111 @@ private List<ProfileReviewActivityDTO> buildReviewActivity(Long userId, Long wor
...
+    }
+
+    private boolean matchesRepository(PullRequestInfoDTO pullRequest, Set<Long> repositoryIds) {
+        Long repositoryId = pullRequest.repository() != null ? pullRequest.repository().id() : null;
+        return matchesRepository(repositoryId, repositoryIds);
+    }
+
+    private boolean matchesRepository(Long repositoryId, Set<Long> repositoryIds) {
+        return repositoryIds == null || repositoryIds.isEmpty() || repositoryIds.contains(repositoryId);
+    }
```
FelixTJDietrich, line 339: Two small ones:

- `limitList` is a one-liner — inline `items.stream().limit(n).toList()` at the two call sites.
- The three `matchesRepository` overloads collapse to one `Set.contains` lambda per filter site:

```java
.filter(a -> resolvedRepositoryIds.isEmpty()
    || resolvedRepositoryIds.contains(a.pullRequest().repository().id()))
```

webapp/src/components/profile/ProfileContent.tsx
```diff
@@ -170,8 +250,95 @@ export function ProfileContent({
...
 	);
 }
+
+interface ActivityMonitorConfigurationProps {
+	repositories: RepositoryInfo[];
+	filters: ActivityMonitorFilters;
+	onRepositoryChange: (repositoryId: number, checked: boolean) => void;
+}
+
+function ActivityMonitorConfiguration({
```
FelixTJDietrich: `ProfileTimeframePicker`, `ReviewActivityCard`, `ProfileHeader` all live in their own files. Move `ActivityMonitorConfiguration` + `ActivityMonitorCheckbox` into `ActivityMonitorConfiguration.tsx` to match.

webapp/src/routes/_authenticated/w/$workspaceSlug/user/$username/index.tsx
```diff
@@ -1,7 +1,12 @@
 import { useQuery } from "@tanstack/react-query";
 import { createFileRoute, retainSearchParams, useNavigate } from "@tanstack/react-router";
 import { z } from "zod";
-import { getUserProfileOptions, getWorkspaceOptions } from "@/api/@tanstack/react-query.gen";
+import {
+	getActivityMonitorOptions,
+	getUserProfileOptions,
+	getWorkspaceOptions,
+} from "@/api/@tanstack/react-query.gen";
+import type { ActivityMonitorFilters } from "@/components/profile/ProfileContent";
```
FelixTJDietrich: Move `ActivityMonitorFilters` to `webapp/src/lib/` (or inline it here). A route importing a type from a component file is a layering inversion.

FelixTJDietrich (commented):
Two things I'd fix (placeholder zeros in the stats DTO and missing tests on the new path), two worth a discussion (loading state, dual-source props), four optional cleanups.

Bigger design question: if `getUserProfile` took `repositoryIds` + `limit`, you wouldn't need `getActivityMonitor` — and that collapses the dual-source props and the placeholder zeros. Any reason for the split?

Not raising (pre-existing on main): `ActivityBadges` prop count, the skeleton type cast, the dual fallback in `reviewedPullRequestsForPopover`, the `Boolean(username/workspaceSlug)` guards.

_Prepared with Claude Opus 4.7._

FelixTJDietrich:
I'll let CC do a final review loop and implement the changes. There was also some issues I introduced through the dependency update.

server/src/main/java/de/tum/cit/aet/hephaestus/profile/UserProfileService.java
FelixTJDietrich: Addressed in abf8a221a — `buildMonitorActivityStats` is gone; `getActivityMonitor` now reuses `ProfileActivityQueryService.getActivityStats` for the badges (real workspace-wide counters) while the lists stay repo-filtered. Stats DTO carries truthful numbers across the wire.

server/src/test/java/de/tum/cit/aet/hephaestus/profile/UserProfileServiceTest.java
FelixTJDietrich: Addressed in abf8a221a — added a `getActivityMonitor` nested test class (5 cases: null workspace, unknown login, repo filter, limit clamp at null/0/9999, repo dedup) and migrated the 8 existing CQRS hydration tests over to `getActivityMonitor` since that's the only caller of `buildReviewActivity` after the split. 2863 unit tests green.

webapp/src/routes/_authenticated/w/$workspaceSlug/user/$username/index.tsx
FelixTJDietrich: Addressed in abf8a221a — added `(activityMonitorQuery.isPending && !activityMonitorQuery.data)` to the `isLoading` OR. First paint now shows skeletons; refetches keep the previous data via `placeholderData`.

webapp/src/components/profile/ProfileContent.tsx
FelixTJDietrich: Addressed in abf8a221a — went the whole way and split `ProfileDTO` into a header (identity / league / contributed repos / XP) while moving the time-windowed activity (review activity, open PRs, stats, reviewed PRs) onto `getActivityMonitor` exclusively. Legacy props dropped from `ProfileContent` and `ProfilePage`; the dual-source fallbacks are gone. Header doesn't refetch when filters change.

server/openapi.yaml
FelixTJDietrich: Addressed in abf8a221a — added `@Validated` to the controller plus `@Min(1) @Max(100)` on `limit` and `@Size(max = 100)` on `repositoryIds`. `openapi.yaml` now emits `minimum`, `maximum`, and `maxItems`; the service-side clamp stays as defense in depth.

server/src/main/java/de/tum/cit/aet/hephaestus/profile/UserProfileService.java
FelixTJDietrich: Addressed in abf8a221a — `limitList` inlined, three `matchesRepository` overloads collapsed to one `filterByRepository` helper, `normalizeRepositoryIds` removed, `collectMonitorRepositories` rewritten as a single `LinkedHashMap` pass.

webapp/src/components/profile/ProfileContent.tsx
FelixTJDietrich: Addressed in abf8a221a — extracted to `webapp/src/components/profile/ActivityMonitorConfiguration.tsx`, with the checkbox folded inline.

webapp/src/routes/_authenticated/w/$workspaceSlug/user/$username/index.tsx
FelixTJDietrich: Addressed in abf8a221a — `ActivityMonitorFilters` and the `DEFAULT_ACTIVITY_MONITOR_LIMIT` / `MAX_ACTIVITY_MONITOR_LIMIT` constants live in `webapp/src/lib/activity-monitor.ts` now.

FelixTJDietrich:
Pushed abf8a221a to address the 8 review threads in one go (and threads are resolved).

Highlights:

- Backend: `ProfileDTO` is now just the header (identity / league / contributed repos / XP); all the time-windowed activity lives on `getActivityMonitor`, which reuses `ProfileActivityQueryService.getActivityStats` for real badge counts. No more hard-coded zeros, no dual-source props, helper sprawl collapsed.
- Controller: `@Validated` + `@Min/@Max/@Size` so `openapi.yaml` documents the contract; service-side clamp kept as defense in depth.
- Tests: 5 new `getActivityMonitor` cases, 8 existing CQRS tests migrated over since `buildReviewActivity` is only reached through that call now. 2863 unit + arch tests green.
- Frontend: `ActivityMonitorFilters` + limit constants moved to `webapp/src/lib/activity-monitor.ts`; `ActivityMonitorConfiguration` extracted; `ProfileContent` reads only from `activityMonitorData` and no longer casts skeletons; `activityMonitorQuery.isPending` now feeds `isLoading` so first paint shows skeletons.

Net: +582 / -971 LOC. Let me know if anything looks off.

_Prepared with Claude Opus 4.7. Human signed off before pushing._

FelixTJDietrich:
Pushed b7300d774 — sync hardening (out of scope for the monitor itself, but needed so a clean-state ls1intum sync is fast and error-free). Verified end-to-end on a clean Postgres with GitHub App auth, plus the staging NATS webhook stream.

**Disable project + discussion sync by default**
- New `hephaestus.sync.projects.enabled` / `hephaestus.sync.discussions.enabled` flags (default `false`). These are the slowest parts of a sync (projects walk projects→fields→status→items→field-values; discussions paginate discussions→comments→replies) and aren't consumed yet.
- Gated on **every** path so the flag is coherent: startup/polling sync, scheduled sync + orphan-item relink, embedded project items during issue/PR sync, and the webhook NATS handlers (project/discussion events route to "no handler" → acked + skipped).

**Fix GraphQL 256 KB buffer limit** (`DataBufferLimitException ... 262144`)
- The GraphQL WebClients register a custom `JacksonJsonDecoder` (for GitHub's 64-bit `databaseId`). `defaultCodecs().maxInMemorySize(16MB)` does **not** apply to custom-registered codecs, so the decoder kept Spring's 256 KB default and large PR pages (embedded reviews/threads/comments) failed and retried forever. Now set on the custom decoder in both `GitHubGraphQlConfig` and `GitLabGraphQlConfig`.

**Verified**
- Incremental + backfill of `ls1intum/Hephaestus`: status COMPLETED, **0 buffer errors**; projects=0, discussions=0 (skipped). Backfill paginates historical issues/PRs with resumable checkpoints.
- Staging NATS replay: PR/issue/push/review events processed; project/discussion events skipped.
- 2863 unit + architecture tests pass.

**One pre-existing issue I did _not_ fix (flagging, not fixing — unrelated subsystem):** under concurrent activity-event processing during sync, `AchievementEventListener` occasionally logs `ObjectOptimisticLockingFailureException`. Root cause: `checkAndUnlock` has `@Retryable(ObjectOptimisticLockingFailureException)`, but the exception is thrown at transaction **commit** — outside the retry interceptor's scope in Spring's `org.springframework.resilience` model — so it escapes. It's caught, logged, non-blocking, and self-heals on the next event/recalc. Fixing it properly means reworking retry/transaction interceptor ordering (global blast radius), so it deserves its own PR + tests rather than riding along here.

_AI-assisted (Claude Opus 4.8, 1M context). Human reviewed before pushing._

Issue #1204, opened 2026-05-18
epic: Slack mentor parity and configurable proactive reflection


FelixTJDietrich:
## Why this matters

The Hephaestus mentor is only reachable from the webapp today. For teams that live in Slack, the friction of context-switching reduces engagement with formative-feedback turns to near zero. We need to expose the mentor over Slack DM at functional parity with the in-app surface, and we need a configurable proactive-reflection cadence — but not at the cost of turning Hephaestus into the bot that workspaces mute.

## Current state

- Mentor is webapp-only; the orchestrator emits `UIMessageChunk` events but there is no Slack rendering path.
- Slack is wired only for outbound leaderboard messages.
- No proactive reflection cadence exists.
- DM safety primitives (kill switch, per-workspace daily cap, dry-run, shared budget) do not exist.

## Proposed

Implement a `SlackChannelAdapter` that translates the mentor orchestrator's `UIMessageChunk` events into Block Kit messages, posted via a per-workspace Bolt `App`. Add three entry points — App Home tab CTA, `/hephaestus mentor` slash command, App Home open — that resume or start a `mentor_session` of surface `SLACK_DM`. Add a `MentorReflectionScheduler` that runs reflection sessions per a user-configured cadence (`OFF` / `WEEKLY` / `END_OF_SPRINT` / `DAILY`, default WEEKLY Monday 09:00 in the user's timezone). All proactive sends pass through stacked safety rails: per-user min-interval (24h, 18h for DAILY), per-workspace daily DM cap (default 50), a Spring-property kill switch, and a dry-run mode that routes to the workspace admin instead of the user. A shared notification budget reconciles mentor reflection DMs with finding-delivery DMs so a learner never wakes up to four bot messages at once.

Identity linkage is a hard prerequisite: a user without a linked Slack identity gets a capped, opt-out-able nag flow (max 2 nags ever, with a "stop reminding me" button that writes a permanent suppression row).

## Acceptance criteria

- [ ] A user with a linked Slack identity opens a mentor DM via App Home, slash command, and App Home CTA; output matches the web mentor's chunk sequence rendered as Block Kit.
- [ ] `MentorReflectionScheduler` runs reflection sessions on the user's configured cadence; integration test seeds three users with `WEEKLY`, `END_OF_SPRINT`, `OFF` and asserts only the first two receive a DM at the right window.
[...]

Issue #1200, opened 2026-05-18
epic: identity-link framework and Sign in with Slack


FelixTJDietrich:
## Why this matters

Hephaestus needs to attribute activity that arrives from Slack, GitHub, GitLab, and Outline to a single user account. Today there is no primitive for binding a Hephaestus account to a verified external identity. Slack delivery in particular cannot route "DM the author of this MR" without that binding. Doing this safely is harder than it looks — email-based auto-matching is a well-documented account-takeover vector, and per-source ad-hoc tables would calcify into N exception cases.

## Current state

- No `identity_link` table, no SPI, no link UI.
- Each provider's user identity is implicitly assumed to match by email or username; this is unsafe and produces silent mis-attribution.
- A workspace admin reinstalling the Slack integration would invalidate any informal mapping.

## Proposed

Add an `identity_link(user_id, provider, team_id, external_id, linked_at, linked_via)` table with a unique `(provider, team_id, external_id)` index and a non-unique `(user_id, provider)` index. Introduce an `IdentityLinkProvider` SPI under `workspace/identity/spi/`, with Slack-OIDC as the first concrete implementation. Add a Sign-in-with-Slack flow that the user explicitly initiates from account settings (never auto-linked by email or handle similarity). Slack OIDC is a *link* flow, not a login replacement. Token rotation lives in the SPI; integration reinstall does not invalidate links because the link binds to `(provider, team_id, external_id)`, not to a workspace integration row. Re-binding an external identity to a new user requires explicit unlink plus a fresh OIDC roundtrip plus two audit-log rows.

The same primitive will host Outline, GitHub, and GitLab identities later; this issue ships the SPI, lifecycle, and the Slack concrete implementation.

## Acceptance criteria

- [ ] `identity_link(user_id, provider, team_id, external_id, linked_at, linked_via)` exists with unique `(provider, team_id, external_id)` and non-unique `(user_id, provider)` indexes.
- [ ] `IdentityLinkProvider` SPI exposes `findUserByExternal(provider, team_id, external_id)` and `findExternalsForUser(user_id, provider)`. Slack delivery code consumes the SPI, not the table.
- [ ] A user completes Sign-in-with-Slack from account settings. Success writes one `identity_link` row; failure surfaces a structured error with no partial state.
- [ ] Reinstalling a workspace's Slack integration (delete + recreate the `workspace_integration` row) leaves existing `identity_link` rows intact. Integration test asserts this.
[...]

PR #1341, opened 2026-07-03
feat(integration): bring the practice mentor to Slack

FelixTJDietrich:
## Description

Brings the AI practice mentor to Slack on two surfaces: developers can DM the app for the same practice feedback they get on pull requests and issues (streamed via Slack's chat streaming API), and workspace admins can opt specific channels into conversation-practice monitoring — consent-gated, forward-only, and erasable end to end.

Part of #1204 (the mentor half; the proactive-reflection surface in that epic is deliberately out of scope). Relates to #1200 — Slack ships as a link-only identity provider here.

## What's in this PR

**Mentor DMs.** A developer links their Slack identity (Sign-in-with-Slack, link-only — no account creation from Slack), then DMs the app. Mentor turns stream as one message that grows (`chat.startStream`/`appendStream`/`stopStream`), with idempotent turn creation, one in-flight turn per thread (DB-enforced), and Retry-After-aware backoff throughout.

The boot-time default-model seeder now accepts an OpenAI-compatible base URL (`AGENT_DEFAULT_CONFIG_BASE_URL`), so instances running self-hosted models can seed a working mentor without manual admin setup.

**Channel monitoring, consent-first.** Nothing is read by default:

- A channel enters the allow-list `PENDING` (bot invited, or added by an admin) and only an admin transition makes it `ACTIVE`: activation posts a visible in-channel announcement and stamps the forward-only boundary — only messages after the announcement are ingested, never history. `ACTIVE ⇄ PAUSED` keeps data; `REVOKED` erases it. Every transition writes an immutable audit row.
- Any member can opt out from the App Home with one click. Opt-out stops ingestion of their messages **and erases** what was already collected — raw messages, thread participation, and the derived conversation observations/feedback about them. Members joining an already-active channel get a just-in-time ephemeral notice with the same one-click opt-out.
- Author deletions and edits in Slack propagate to the stored copy — deletions in every consent state, including `PAUSED`.
- Erasure reaches raw **and derived** data on all five triggers: person opt-out, channel revoke, retention sweep (thread-grain), app uninstall, workspace purge. `tokens_revoked` distinguishes bot-token revocation (teardown) from a member revoking their personal sign-in (no-op), so one person's revocation cannot destroy a workspace's data. The Slack JetStream transport buffer expires after 72h so message text never outlives erasure in the broker; the DSMS record of processing is updated accordingly.

**Slack-side lifecycle awareness.** The app subscribes to channel lifecycle events and reacts so consent state can never silently diverge from Slack reality: the bot being removed or the channel archived pauses monitoring (audited, admin-reversible), deletion in Slack revokes and erases our copy, renames heal the stored name. Uploads with a comment (`file_share`) are ingested for their text — the file itself is never persisted.

**Nightly reconciliation, at SCM-sync parity.** Slack retries a dropped event only three times within ~6 minutes, so a longer receiver outage would silently lose messages. A nightly scheduler (cron + cross-pod lock, like the SCM sync) refreshes channel metadata (pausing ACTIVE channels only on definitive signals — never on a transport blip) and replays `conversations.history` through the **same consent-gated ingest stack** the event path uses: the fetch floor is `max(consent announcement, retention cutoff, last watermark)`, consent is re-read before every request, resuming a paused channel stamps the watermark so a paused gap is never backfilled, and the sync never infers deletions from absence. A workspace with no active Slack connection is skipped before any call is attempted, so it cannot burn the budget on doomed requests. Budgets and 60s self-pacing are designed for Slack's 2025 non-Marketplace rate clamp (1 request/min, 15 objects on history/replies) and configurable upward for Marketplace/internal installs.

**Durable ingestion on the existing transport.** Channel messages ride the same `webhook → JetStream → consumer` pipeline as GitHub/GitLab: HMAC-verified (constant-time, 300s replay window, raw-body), deduplicated by `event_id`, idempotent on redelivery (`INSERT … ON CONFLICT`), 503 on publisher unavailability so Slack redelivers. A publish-side gate drops non-consented channel content before it is ever persisted to the stream. Interactive events (mentor turns, App Home, opt-out buttons) are handled with the consent/erasure state change completing before the ack and Slack follow-up calls running async, keeping the 3-second ack budget independent of Slack API latency.

**Detection.** Quiescent consented threads become `CONVERSATION_THREAD` review jobs, and the resulting feedback is delivered back to participants — gated by the same consent checks at read time, with untrusted channel text quarantined from the instruction layer. Conversation reviews mount the same cross-artifact workspace inventory issue reviews get (aggregated across monitored repositories — a conversation isn't anchored to one), so every job type sees a consistent context surface; only clone/diff-dependent sources are excluded.

**Admin UI.** Workspace **Admin → Settings** gains a channel table (status, opted-out count, audit history sheet) with a searchable channel picker. Activation is an explicit confirm dialog enumerating consequences; removal is a type-the-channel-ID destructive dialog (relaxed to a plain confirm for `PENDING` channels that never collected anything). Query failures render as errors with retry — never as an empty "nothing monitored" state.

**Schema & boundaries.** One consolidated Liquibase changelog (Slack tables with workspace FKs, history-sync watermarks, `chat_thread` surface, `feedback_placement`, the `git_provider → identity_provider` rename + SLACK seed, CHECK widening for `CONVERSATION_THREAD`), preconditioned with rollbacks, migrating clean from empty. The mentor `agent` module reaches Slack data only through agent-owned SPIs (ArchUnit-enforced), and the core consumer stays vendor-neutral. Every query lives in a Spring Data repository (no raw `JdbcTemplate` outside the documented tenant resolver — enforced). Additional architecture tests pin the module's own invariants: manifest `bot_events` ↔ registered handler parity, every Slack controller workspace-scoped or a hidden receiver, every NATS consumer on the shared handler bases, repository finders workspace-scoped at compile time, stereotype-promising class names (`*Service`/`*Repository`/`*Controller`) carrying their stereotype, no own-package fully-qualified names in code bodies, and Slack-prefixed operationIds in the OpenAPI spec. `openapi.yaml`, the webapp client, and the ERD are regenerated.

## Scope notes

- The interactive mentor path stays in-process (a dropped reply is user-recoverable and needs low-latency streaming); only passive channel content rides the durable transport.
- The message-safety guard is a deterministic keyword fast-path behind a pluggable seam — it is not crisis detection and is named accordingly.
- The `git_provider → identity_provider` rename is DB-level; renaming the pre-existing Java SPI identifiers (26 shared auth files) is deferred to keep this PR reviewable.
- Full-history reconciliation of the mentor DM channels is deferred: rehydrating missed user DMs into the mentor's chat threads needs a write port into the mentor module, and automatically re-driving a mentor turn days later risks double replies. Missed-event healing currently covers monitored channels only.
- Bot-authored messages in consented channels are deliberately not ingested (data minimization); `channel_id_changed` (Slack Connect conversions) is not handled — both are documented tradeoffs.
- `channel_unarchive` is deliberately not subscribed: resuming monitoring is an admin-only consent decision, never an automatic reaction.
- A test deployment runs this branch end to end: the signed-webhook path was verified live (valid signature accepted, expired timestamp / tampered body / bogus signature all rejected with 401), the schema migrated from an empty database, and the nightly reconciliation was driven against real data to confirm its fail-safe invariants. The Slack OAuth install and DM/channel flows still need a workspace installation to exercise; the suite below covers everything deterministic.

## How to test

```bash
# server
mvn test                        # unit tier
mvn test -Parchitecture-tests   # Modulith + tenancy + agent↔slack boundary + vendor-neutral consumer + Slack invariants
mvn verify                      # + integration tier (Testcontainers): consent lifecycle E2E, ingest
                                #   idempotency/redelivery, erasure across all five triggers, purge isolation

# webapp
pnpm run check && pnpm run test:webapp && pnpm --filter webapp run build-storybook
```

Key suites: `SlackConsentLifecycleE2EIntegrationTest` (register → activate → ingest → opt-out → revoke, asserted at every hop), `SlackChannelMessageHandlerTest` (transient failure → redelivery → applied exactly once), `SlackChannelHistorySyncServiceTest` (fetch-floor arithmetic, consent re-checks, never-tombstones, watermark discipline), `SlackAppHomeOptOutErasureIntegrationTest` / `SlackRetentionErasureIntegrationTest` / `WorkspacePurgeIntegrationTest` (erasure reaches derived rows, never over-reaches), `SlackSignatureVerifierTest` (replay window), `SlackManifestTemplateTest` (manifest ↔ handler parity).

**Live:** a test instance of this branch is deployed (signed-webhook verification and reconciliation invariants already exercised against it). Full flow: install via the manifest template (`docs/admin/slack-app-manifest-template.yml`) → activate a channel (announcement posts) → post a message → confirm it is ingested → opt out as a participant (rows erased) → revoke (erased, audit survives) → DM the app for a streamed mentor turn.

## Screenshots

The admin surface is covered by Storybook stories (`AdminSlackChannelsSettings`: default / empty / error / activation dialog / remove dialog / picker; `SettingsPage` with Slack preferences). Screenshots from the test workspace can be added on request.

## Notes for reviewers

66 commits, so read the diff rather than the history. Suggested reading order: `integration/slack/events` (verification + ingest) → `webhook` (handlers + publish gate) → `channel` (consent state machine + admin API) → `sync` (nightly reconciliation) → `agent/handler/conversation` (detection + delivery) → webapp `admin/slack-channels`. Squash-merge recommended; the PR title is the intended release commit.


server/src/main/java/de/tum/cit/aet/hephaestus/agent/handler/FeedbackLedgerRecorder.java
```diff
@@ -68,24 +68,35 @@ public class FeedbackLedgerRecorder {
...
+     * positions from the one shared constant rather than a second literal.
+     */
+    public static final int CONVERSATION_UNIT_ORDINAL_BASE = 3000;
+
     private final ObservationRepository observationRepository;
     private final FeedbackRepository feedbackRepository;
     private final FeedbackObservationRepository feedbackObservationRepository;
     private final FeedbackPlacementRepository feedbackPlacementRepository;
     private final PracticeReviewProperties reviewProperties;
+    private final org.springframework.context.ApplicationEventPublisher eventPublisher;
```
FelixTJDietrich: Please use import! Do a audit across the codebase and fix all occurences!

server/src/main/java/de/tum/cit/aet/hephaestus/core/tenancy/WorkspaceScopedTables.java
```diff
@@ -69,6 +69,9 @@ public class WorkspaceScopedTables {
         "worker_registry",
         // Instance-scoped OAuth login providers (sign-in options); shared across all workspaces
         "login_provider",
+        // Fleet-wide Slack mentor daily LLM-spend budget counter, keyed by UTC day — a single global budget
+        // across all replicas, not workspace-scoped; carries no workspace content.
+        "slack_mentor_daily_budget",
```
FelixTJDietrich: What is up with that? I think we don't need this? Remove the whole budget stuff?! We maybe want to do something in a follow-up but for now we want to just analyze the practices of activity events as they come in. Pressure test and cleanup completely, do a full sweep. Search for similar things that look off. We want a cohesive practice detection architecture!!! This is fatal

server/src/main/java/de/tum/cit/aet/hephaestus/integration/core/webhook/SlackNatsPublisherConfiguration.java
```diff
@@ -0,0 +1,66 @@
...
+ *
+ * <p>Requires the {@code natsConnection} bean ({@code hephaestus.sync.nats.enabled=true}); if
+ * NATS is disabled the publisher is simply absent and the Slack events endpoint replies 503 on
+ * the channel branch (Slack redelivers) rather than dropping content silently. The unauthenticated
+ * {@code /slack/*} payload-size guard is NOT registered here — it lives in
+ * {@code SlackWebSecurityConfiguration} (gated only on {@code slack.enabled}) so it is present even
+ * in this NATS-off topology where this config does not load.
+ *
+ * @see WebhookConfiguration the webhook-role owner of the same {@link WebhookProducerBeans}
+ */
```
FelixTJDietrich: We have lots of long comments in this codebase and the changes, please go through all of them and understand if we really follow absolute best practices when it comes to documentation. We want to cut through unecessary bloay, theatre and keep them evergreen and truthful. A lying comment is bad. Also look at all the package-info.java and so on. Make sure to use subagents to pressure test the comments and snowball.

server/src/main/java/de/tum/cit/aet/hephaestus/integration/core/webhook/SlackWebSecurityConfiguration.java
```diff
@@ -0,0 +1,44 @@
...
+    @Bean
+    FilterRegistrationBean<WebhookPayloadSizeFilter> slackPayloadSizeFilter(
+        WebhookProperties properties,
+        MeterRegistry meterRegistry
+    ) {
+        FilterRegistrationBean<WebhookPayloadSizeFilter> registration = new FilterRegistrationBean<>(
+            new WebhookPayloadSizeFilter(properties, meterRegistry)
+        );
+        // Covers both /slack/events and /slack/interactivity. HIGHEST_PRECEDENCE so the Content-Length cap
+        // runs before Spring buffers the body and before the signature verifier.
```
FelixTJDietrich: Excuse me?! I though `/slack/events` and `/slack/interactivity` are tech debt?! Please cut through all tech dept in this branch! We want to use a cohesive and consistent approach! Really audit the full changes, leverage diff compared to main and cut through all the bullshit. Look at other code across the codebase and be architecturally consistent, you might need to refactor and change architecture tests and so on. Leverage subagents

server/src/main/java/de/tum/cit/aet/hephaestus/integration/core/webhook/WebhookJetStreamBootstrap.java
FelixTJDietrich, line 1: Wait what is this streaming setup?! Pressure test, I though we are just using NATS as a message queue. Streaming to the client or to slack should probably be done directly?! But sometimes a redis is recommended? Not sure, I mean if you want to continue the stream where it left off.

server/src/main/java/de/tum/cit/aet/hephaestus/integration/scm/github/project/GitHubProjectStatusUpdateMessageHandler.java
```diff
@@ -49,14 +49,14 @@ public class GitHubProjectStatusUpdateMessageHandler
     private final ProjectRepository projectRepository;
     private final GitHubProjectStatusUpdateProcessor statusUpdateProcessor;
     private final ScopeIdResolver scopeIdResolver;
-    private final GitProviderRepository gitProviderRepository;
+    private final IdentityProviderRepository gitProviderRepository;
```
FelixTJDietrich, line 52: I like the whole naming improvements of the PR. Please understand them and the theme. We wanted to make everything more human friendly, understandable and cohesive / consistent. Please snowball even more. Pressure test using subagents!

server/src/main/java/de/tum/cit/aet/hephaestus/integration/slack/channel/SlackChannelAdminController.java
```diff
@@ -0,0 +1,112 @@
...
+ * {@code (workspace_id, slack_channel_id)}. Every method scopes on the {@link WorkspaceContext} workspace id, so a
+ * channel of another workspace resolves to 404 (isolation). Illegal transitions surface as {@code 409 ProblemDetail}
+ * through {@link SlackChannelControllerAdvice}; not-found / validation / auth flow through the shared advice chain.
+ */
+@WorkspaceScopedController
+@RequestMapping("/slack/channels")
+@RequireAtLeastWorkspaceAdmin
+@ConditionalOnProperty(name = "hephaestus.integration.slack.enabled", havingValue = "true", matchIfMissing = false)
+@Validated
+@Tag(name = "Slack channel activation", description = "Per-workspace Slack channel consent + activation")
```
FelixTJDietrich, line 46: Really make sure that we are doing this stuff consistently and correctly, this also includes the access controls and so on. Use subagents to cover all changes !

server/src/main/java/de/tum/cit/aet/hephaestus/integration/slack/connect/SlackConnectionStrategy.java
```diff
@@ -34,12 +34,35 @@ public class SlackConnectionStrategy implements ConnectionStrategy {
...
+        // Delivery of `message` events with channel_type=im — the DM that drives a mentor turn
+        // (SlackEventDispatcher). Without it the mentor DM entry point receives nothing.
+        "im:history",
+        // team/workspace metadata resolution.
         "team:read",
+        // users.list + Slack user → workspace member identity resolution (SlackMentorIdentityResolver).
         "users:read",
+        // email-keyed member matching when linking a Slack user to a Hephaestus member.
         "users:read.email"
     );
```
FelixTJDietrich, line 49: Also document in the project / contributor docs, keep consistent!

server/src/main/java/de/tum/cit/aet/hephaestus/integration/slack/conversation/SlackConversationProjector.java
FelixTJDietrich, line 1: Think about all of this and check out if this is tech debt. Really deeply audit for tech debt in this PR. Critically challenge your requirements and do a principled approach. Work from first principles and challenge your assumptions and so on. We wont to converge on the best possible longterm solution that works well and simple and is maintainable

server/src/main/java/de/tum/cit/aet/hephaestus/integration/slack/events/ObviousAbuseFastPathSlackSafetyClassifier.java
FelixTJDietrich, line 1: Can we do even better? Critically challenge legerage subagents. Also compare with the events stuff of other integrations. I really need you to deeply compare with the scm integration and understand similarities and differences and then converge on the best possible solutin

server/src/main/java/de/tum/cit/aet/hephaestus/agent/mentor/chat/wire/TranslatorState.java
```diff
@@ -244,6 +144,15 @@ public synchronized void recordDataFinding(UUID findingId) {
...
         partsAccumulator.add(part);
+        linkedFindingIds.add(findingId);
+    }
+
+    /**
+     * The observation ids the mentor linked this turn via {@code link_finding}, in emission order (duplicates
+     * retained - the reconciler de-duplicates). Snapshot copy for cross-thread safety on the finalise path.
+     */
+    public synchronized java.util.List<UUID> linkedFindingIds() {
+        return java.util.List.copyOf(linkedFindingIds);
```
FelixTJDietrich: Here you are not importing again, do a full sweep

server/src/main/java/de/tum/cit/aet/hephaestus/agent/mentor/chat/MentorChatService.java
FelixTJDietrich, line 1: I think the prompt management could be much better and idiomatic. This looks horrible. Pressure test and consider alternatives

server/src/main/java/de/tum/cit/aet/hephaestus/agent/AgentJobType.java
```diff
@@ -9,4 +9,10 @@
 public enum AgentJobType {
     PULL_REQUEST_REVIEW,
     ISSUE_REVIEW,
+    /**
+     * Detection over a settled Slack conversation thread. Repo-less: the case context is the
+     * thread's human turns (materialised as {@code inputs/context/conversation_thread.json}), with no
+     * clone, no diff, and no SCM source mount. Handled by {@code ConversationReviewHandler}.
+     */
+    CONVERSATION_REVIEW,
```
FelixTJDietrich: No here it should also have full context and access to all the prs issues and repo. We always want to have  similar context per job to keep it maintainable and powerful. Please ensure it is consistent and cohesive. The comment is also not necessary

server/src/main/java/de/tum/cit/aet/hephaestus/integration/core/framework/IntegrationFrameworkBootstrap.java
```diff
@@ -48,23 +48,24 @@ public class IntegrationFrameworkBootstrap {
     private final List<InlineFindingChannel> inlineFindingChannels;
     private final List<ApprovalChannel> approvalChannels;
     private final List<IntegrationLifecycleListener> lifecycleListeners;
+    private final boolean webhookRoleEnabled;
 
     public IntegrationFrameworkBootstrap(
         IntegrationManifestRegistry manifests,
         List<WebhookSignatureVerifier> signatureVerifiers,
-        List<WebhookSecretSource> secretSources,
+        @SuppressWarnings("unused") List<WebhookSecretSource> secretSources,
```
FelixTJDietrich: This seems to be a bandaid? Please sweep for similar bandaids and cut through them

server/src/main/java/de/tum/cit/aet/hephaestus/integration/slack/conversation/SlackConversationCandidateSource.java
```diff
@@ -0,0 +1,147 @@
...
+            rs -> {
+                long[] participants = readLongArray(rs.getArray("participant_member_ids"));
+                out.add(
+                    new ConversationThreadCandidate(
+                        rs.getLong("workspace_id"),
+                        rs.getLong("id"),
+                        rs.getString("slack_channel_id"),
+                        rs.getString("slack_thread_ts"),
+                        rs.getString("last_ts"),
+                        rs.getString("last_reviewed_ts"),
```
FelixTJDietrich: Is this correct? Please audit all components and critically challenge if they should be services and also this is a database query which should probably be in a repository. I really need you to do a deep sweep across all with subagents and fix all those hacky bandaid solutions. We need a clean architecture and Component is probably not helping here. Think about architecture tests and also codebase consistency and cohesion. You really need to not miss any of those issues, do not defer. Please fix in this PR!

server/src/main/java/de/tum/cit/aet/hephaestus/integration/slack/conversation/SlackConversationProjector.java
```diff
@@ -0,0 +1,217 @@
...
+            """,
+            rs -> {
+                keys.add(
+                    new ThreadKey(
+                        rs.getString("slack_channel_id"),
+                        rs.getString("channel_name"),
+                        rs.getString("slack_thread_ts"),
+                        rs.getLong("message_count")
+                    )
+                );
```
FelixTJDietrich: Same here?! Why this query, this is a bandaid!

server/src/main/java/de/tum/cit/aet/hephaestus/integration/slack/conversation/SlackConversationProjector.java
FelixTJDietrich, line 1: Critically challenge this file architecturally and snowball into similar files if this is really how it should be.

server/src/main/java/de/tum/cit/aet/hephaestus/integration/slack/conversation/SlackConversationSourceLiveness.java
```diff
@@ -0,0 +1,60 @@
...
+                "SELECT t.id FROM slack_thread t " +
+                    "JOIN slack_monitored_channel c " +
+                    "  ON c.workspace_id = t.workspace_id AND c.slack_channel_id = t.slack_channel_id " +
+                    "WHERE t.workspace_id = ? AND t.id IN (" +
+                    placeholders +
+                    ") AND c.consent_state = 'ACTIVE'",
+                Long.class,
+                args
+            )
+        );
```
FelixTJDietrich: Here too, make sure to have a clean architecture across all files. Please work through all the server files and ensure this is not a thing. Think about file name endings that are allowed, I thought we want Services to have a Service suffix and Repositories a Repository one. Research using websearch and find trusted sources of industry and OSS best practices. Cite evidence and bridge to A+!

server/src/main/java/de/tum/cit/aet/hephaestus/integration/slack/domain/SlackChannelConsentEvent.java
FelixTJDietrich, line 1: For all the JPA schema classes please pressure test each column. Adversarially challenge and consider alternatives. Make sure all is following absolute best practices.

server/src/main/java/de/tum/cit/aet/hephaestus/agent/handler/FeedbackLedgerRecorder.java
FelixTJDietrich: Swept the whole codebase: every inline fully-qualified name is now an import (JPQL enum literals and genuine name collisions excepted). A new source-scan architecture test (`OwnPackageImportArchTest`) bans own-package FQNs in code bodies permanently — it already caught seven stragglers in `PracticeDetectionResultParser` during the sweep. Fixed in bd40ea962.

server/src/main/java/de/tum/cit/aet/hephaestus/core/tenancy/WorkspaceScopedTables.java
FelixTJDietrich: The budget subsystem was fully removed in a5a5949e2 — no `slack_mentor_daily_budget` table, entity, guard, or test remains (grep is clean, and `WorkspaceScopedTables` no longer lists it). Practice detection now analyzes activity as it arrives (events + the consent-gated nightly reconciliation), with no quota layer in between. Cost control is one-in-flight-turn-per-thread plus sandbox session caps; a spend quota can come back as its own follow-up if needed.

server/src/main/java/de/tum/cit/aet/hephaestus/integration/core/webhook/SlackNatsPublisherConfiguration.java
FelixTJDietrich: Two comment-audit rounds executed across the diff: process narration (slice tags, plan refs, defect war stories) purged; comments made false by refactors fixed (e.g. the "unmapped array column" and "raw JDBC only" claims); triplicated javadoc essays deduplicated to their canonical home; package-info files trimmed to stable contracts. The standard applied: a comment earns its place only if it states a contract or non-obvious WHY that outlives the next refactor.

server/src/main/java/de/tum/cit/aet/hephaestus/integration/core/webhook/SlackWebSecurityConfiguration.java
FelixTJDietrich: Those legacy paths are gone — this class was deleted with them. Slack ingress is unified on `/webhooks/slack` + `/webhooks/slack/interactivity` behind the same verify→derive→publish pipeline GitHub/GitLab use; `SecurityConfig` permits exactly those two paths and an architecture test pins every Slack controller as workspace-scoped or a hidden receiver.

server/src/main/java/de/tum/cit/aet/hephaestus/integration/core/webhook/WebhookJetStreamBootstrap.java
FelixTJDietrich: Naming collision, not client streaming: a JetStream "stream" is NATS's server-side persistence primitive for the webhook message queue (at-least-once delivery + `Nats-Msg-Id` dedup — this bootstrap just creates one queue store per integration). Streaming to clients never touches NATS: the web mentor streams over SSE directly from the server, and Slack replies stream via Slack's own `chat.startStream`/`appendStream` API. Nothing resumes a byte-stream mid-way, so no Redis is needed — a dropped mentor reply is simply re-sent as a whole turn.

server/src/main/java/de/tum/cit/aet/hephaestus/integration/scm/github/project/GitHubProjectStatusUpdateMessageHandler.java
FelixTJDietrich: Snowballed further: `ContentProvider`→`ContentSource` family, `git_provider`→`identity_provider` (DB + SLACK seed), `WorkspaceAbi`→`SandboxLayout`, the overclaiming safety classifier renamed to the honest `KeywordSlackMentorInputGuard`, DTO/operationId fixes (`getSlackUserPreferences`, `SlackUserWorkspacePreferencesDTO`, `sendSlackTestMessage`), and a directional architecture rule so a name that promises a stereotype (*Service/*Repository/*Controller) keeps that promise.

server/src/main/java/de/tum/cit/aet/hephaestus/integration/slack/channel/SlackChannelAdminController.java
FelixTJDietrich: Audited every Slack controller against the SCM/workspace admin reference set: admin surfaces are `@WorkspaceScopedController` + `@RequireAtLeastWorkspaceAdmin` + server-role gated; receivers are `@Hidden` + webhook-role gated; member-level preferences use the shared `@workspaceSecure.isMember()` pattern. A new architecture rule enforces workspace-scoped-or-hidden for every Slack `@RestController` so the next endpoint can't skip it.

server/src/main/java/de/tum/cit/aet/hephaestus/integration/slack/connect/SlackConnectionStrategy.java
FelixTJDietrich: Documented in `docs/admin/production-setup.mdx`: each requested OAuth scope mapped to the feature that needs it, plus the event-subscription pointer to the manifest template. The scope set itself is pinned by `SlackConnectionScopesTest` and the manifest↔handler parity test.

server/src/main/java/de/tum/cit/aet/hephaestus/integration/slack/conversation/SlackConversationProjector.java
FelixTJDietrich: Reworked from first principles in bd40ea962: the raw-JDBC rationale ("array column unmapped on the entity") had been false since `SlackThread` gained its `@JdbcTypeCode` mapping, so the entire tunnel was obsolete debt. All queries now live in `SlackThreadRepository`/`SlackMessageRepository`; the projector is a thin `@Service` SPI adapter; the schema-contract test was refocused on the remaining native queries.

server/src/main/java/de/tum/cit/aet/hephaestus/integration/slack/events/ObviousAbuseFastPathSlackSafetyClassifier.java
FelixTJDietrich: Converged with the SCM shape: this class was renamed to the honest `KeywordSlackMentorInputGuard` (a deterministic obvious-abuse fast path behind a pluggable seam — not claimed crisis detection), and Slack events now ride the identical webhook→JetStream→consumer pipeline as GitHub/GitLab, with manifest↔handler parity enforced by test and a nightly reconciliation at SCM-sync parity closing the events-only delivery gap.

server/src/main/java/de/tum/cit/aet/hephaestus/integration/slack/mentor/SlackStreamingMentorChannel.java
```diff
@@ -0,0 +1,462 @@
...
+        } catch (Exception e) {
+            // Terminals never throw (contract). A gone recipient just means the stream is already finalized.
+            log.debug("Slack stream finalize skipped (channel={}): {}", channel, e.getMessage());
+        }
+    }
+
+    /** Terminal content write (open or append) with a few transient retries — the flush loop is already stopped. */
+    private void terminalWrite(String text) {
+        int attemptsLeft = 3;
+        while (attemptsLeft-- > 0) {
```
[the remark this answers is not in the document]
FelixTJDietrich: Fixed as suggested: `while (attemptsLeft > 0)` with the decrement explicit in the catch, preserving exactly three attempts and the give-up check.

server/src/main/java/de/tum/cit/aet/hephaestus/agent/mentor/chat/wire/TranslatorState.java
FelixTJDietrich: Fixed in the same sweep — `OwnPackageImportArchTest` now fails the build on any own-package FQN in a code body, so this can't come back.

server/src/main/java/de/tum/cit/aet/hephaestus/agent/mentor/chat/MentorChatService.java
FelixTJDietrich: Prompt assembly moved out of the service: `MentorTurnPromptFactory` owns turn-prompt construction and the per-surface directive lives as a classpath template (`agent/mentor/prompts/slack-dm-turn.md`), mirroring how the system prompt already ships as a resource. The persisted user message stays verbatim; only the runner prompt is decorated, and tests pin that the directive reaches SLACK_DM turns and never WEB ones.

server/src/main/java/de/tum/cit/aet/hephaestus/agent/AgentJobType.java
FelixTJDietrich: Done: CONVERSATION_REVIEW jobs now mount the same cross-artifact workspace inventory ISSUE_REVIEW gets (`WorkspaceInventoryContentSource`, aggregated across every monitored repo since a conversation isn't anchored to one). Only clone/diff-dependent sources stay out, with that boundary stated once where applicability is decided. The enum javadoc is trimmed to one line per value.

server/src/main/java/de/tum/cit/aet/hephaestus/integration/core/framework/IntegrationFrameworkBootstrap.java
FelixTJDietrich: The injected-but-unused `WebhookSecretSource` list is deleted — bean registration never depended on it. Swept the codebase for sibling `@SuppressWarnings("unused")` injection bandaids; the rest were legitimate.

server/src/main/java/de/tum/cit/aet/hephaestus/integration/slack/conversation/SlackConversationCandidateSource.java
FelixTJDietrich: Fixed, not deferred: every query moved into repository methods, the class is a `@Service` SPI adapter, and two new architecture rules make the pattern structural — no `JdbcTemplate` in the Slack module outside the documented tenant resolver, and stereotype-promising names must carry the stereotype. The fleet-wide candidate scan is explicitly allowlisted with its `@WorkspaceAgnostic` caller-chain justification, mirroring the existing enumerators.

server/src/main/java/de/tum/cit/aet/hephaestus/integration/slack/conversation/SlackConversationProjector.java
FelixTJDietrich: Same treatment — the query now lives on the repository (typed projection row), the projector only orchestrates.

server/src/main/java/de/tum/cit/aet/hephaestus/integration/slack/conversation/SlackConversationProjector.java
FelixTJDietrich: Challenged architecturally and reworked: repository-owned queries, thin `@Service` adapter implementing the agent-owned SPI, obsolete raw-JDBC justification comments deleted, and the sibling files (candidate source, liveness) got the identical treatment so the package is uniform.

server/src/main/java/de/tum/cit/aet/hephaestus/integration/slack/conversation/SlackConversationSourceLiveness.java
FelixTJDietrich: Done across the module and enforced: the stereotype rule is directional on purpose — a name ending in Service/Repository/Controller must be that thing (non-brittle, per ArchUnit naming-convention practice), rather than forcing every bean into a fixed suffix zoo. Queries moved to repositories throughout; abstract shared bases are exempted since they make no runtime promise.

server/src/main/java/de/tum/cit/aet/hephaestus/integration/slack/domain/SlackChannelConsentEvent.java
FelixTJDietrich: Pressure-tested every column of all six Slack entities: workspace-first primary key restored on slack_participant_consent (matching every sibling key and the hot opted-out count), workspace FKs added on all six tables, sync watermark columns justified per query pattern, VARCHAR(32) Slack ids kept (Slack says don't assume id length; 32 covers observed formats with headroom), decided_at restamped on every consent re-decision via the ON CONFLICT upserts, and the audit table stays append-only by construction. Changes shipped as preconditioned changesets in the branch changelog; anything speculative was deliberately left out.

PR #1344, opened 2026-07-12
fix: harden practice detection and review execution

FelixTJDietrich:
## Description

Hardens practice detection from review preparation through serving and delivery. No API changes.

- Abort review preparation when Git cannot produce a trustworthy diff instead of allowing a diff-less review to reach the model.
- Select the latest detection run deterministically and within the current workspace, including when runs share a timestamp.
- Exclude artifacts from repositories hidden from contributions across all six aggregate practice-serving queries.
- Seed the default practice catalog for runtime-created workspaces on the bounded application executor, and reconcile every workspace at startup after interrupted or rejected work.
- Make the E2E profile run the PostgreSQL-backed review executor and provide a bounded setup flow for live GitLab and LLM validation.
- Keep sandbox startup reliable on shared Docker hosts by relying on the container-local PID cgroup limit rather than the host-UID-scoped `nproc` limit.
- Prevent Storybook dependency rediscovery reloads by pre-bundling the declared runtime dependency set with discovery disabled; shared Storybook mocks are reset between tests.
- Watch the current pnpm lockfile, workspace definition, and npm configuration in CI path filters and dependency labeling.

## Validation

- 29 sandbox-policy unit tests
- 19 pull-request content-source unit tests
- 17 catalog-seeder unit tests
- 2 E2E-profile overlay tests
- 22 observation-repository integration tests against PostgreSQL
- 1,048 Storybook browser tests
- `pnpm run format`
- `pnpm run check`

A live E2E review against GitLab MR 8 completed and was delivered successfully: PostgreSQL job execution, repository synchronization, checkout and diff generation, Docker sandbox execution, 20 live LLM calls, three persisted observations, and GitLab note publication.


FelixTJDietrich:
Confirmed still needed against main (7dc852afc) — all three fabrication-risk paths are present verbatim, the catalogue seeder is still boot-only, and `ObservationRepository` has no hidden-row handling. Rebase note: the conflicts are a genuine three-hunk content merge in `PullRequestContentSource` (#1400 reworded the same log statements) plus the seeder; the seeder must end up with **both** listeners — main's boot-time `WorkspacesInitializedEvent` and this branch's workspace-created listener — not one replacing the other.

Issue #1380, opened 2026-07-16
ci: validate the full Liquibase changelog chain against real PostgreSQL


FelixTJDietrich:
## Problem

The Liquibase changelog chain **is** the upgrade path — and it is currently untested: the test suite runs against `ddl-auto`, so a broken changeset surfaces only at production boot. This is the single highest-value upgrade guarantee available, and it protects every migration written between now and the release.

## What

- CI job running `liquibase:update` over the full chain against a real PostgreSQL (empty DB → head), on every PR touching `db/changelog/**` and on main.
- Contributor-docs policy: released changesets are never edited; destructive schema changes follow deprecate-then-remove across two releases.

## Priority

**Do now, not at Gate 3** — an afternoon of CI work that de-risks the entire remaining 1.0 schema churn (leaderboard drop #1374, catalog #1361, labels #1365, ledger extensions #1356).

Delivers early, ahead of Gate 3.

Part of #1378.

PR #1390, opened 2026-07-16
ci: enforce Liquibase changelog immutability and document the migration-chain gate

FelixTJDietrich:
## Description

The Liquibase changelog chain is the upgrade path for every self-hosted instance, so a **released changeset must never change** — editing one throws a checksum error at boot on already-upgraded databases, while fresh installs silently get a different schema (#1380).

This PR adds a **`Migrations` quality gate** and documents the guarantees. Deliberately small and honest about scope:

- **New `Migrations` leg** (pure git, no container — runs in ~seconds). It fails a PR when a released file under `db/changelog/` is modified, renamed, or deleted, or when an existing `master.xml` `<include>` line is removed, reordered, or edited. This catches exactly the edits Liquibase **checksums cannot** see: preconditions, contexts, labels, comments, and deleted changesets.
- **The empty→head chain replay already exists.** The `database-validation` leg's `draft-changelog` step already applies the full `master.xml` chain via `liquibase:update` against a real (pg_partman) PostgreSQL and fails CI on a broken chain. Rather than duplicate a Postgres + Maven run, this PR marks that step as load-bearing with a comment.
- **Docs.** Immutability policy + deprecate-then-remove discipline, and an explicit statement of what the replay does *not* cover.

### Why git, not Liquibase checksums

Liquibase excludes preconditions, contexts, labels, and comments from changeset checksums, and silently ignores deleted changesets ([reference](https://www.liquibase.com/blog/what-affects-changeset-checksums)). A checksum-only guard would miss those edits, so immutability is enforced at the git layer against the PR's merge-base.

### Honest limitations (documented in `database-migration.mdx`)

The chain replay runs against an **empty** database via the Liquibase Maven plugin, so it validates that the chain is well-formed and applies — it does **not** reproduce a specific instance's boot (a real `dev`/`prod` boot filters by `spring.liquibase.contexts`), does **not** catch data-incompatible migrations (a `NOT NULL` add or a `CHECK` that existing rows violate), and does **not** exercise `<rollback>` or `MARK_RAN` precondition paths. Those remain a review responsibility via the deprecate-then-remove policy.

## How to test

CI covers this. To verify the new gate on a scratch commit:

- **Immutability (should fail):** change one character inside any existing `server/src/main/resources/db/changelog/*.xml` → the `Migrations` leg fails with *"Released Liquibase changesets were modified…"*.
- **master.xml append-only (should fail):** reorder or edit an existing `<include>` line → fails with *"master.xml is append-only…"*.
- **Normal migration (should pass):** append a new changelog file plus a new `<include>` at the end of `master.xml` → passes.
- **Broken chain (should fail on the `Database` leg):** a non-parsing changeset fails `database-validation` via `liquibase:update`.

<sub>🤖 Reviewed across five principal-engineer passes; the original draft's redundant two-phase Docker+Maven replay was cut after review confirmed the existing `database-validation` leg already covers empty→head.</sub>

## Summary by CodeRabbit

* **New Features**
  * Added a new “database-migrations” CI quality gate, including full migration-chain replay and drift validation.
  * Introduced a changelog immutability guard that fails CI if released Liquibase changelog files are modified/renamed/deleted or if `master.xml` includes are not append-only.

* **Documentation**
  * Updated CI/CD and database migration guidance to explain the new migrations gate, immutability policy, and updated workflow/parallel execution behavior.
  * Refined changelog filename and `master.xml` append instructions for contributors.


FelixTJDietrich:
### Gate verified end-to-end in CI (not just locally)

A passing check is weak evidence for a guard like this — it can pass by doing nothing. Two things were caught and fixed while verifying:

1. **First failure:** `git merge-base origin/main HEAD` exited 128 — `actions/checkout` does not expose `origin/<base>` as a ref. Now the baseline comes from `github.event.pull_request.base.sha`.
2. **A vacuous green:** the gate then "passed" in 7s while the guard step was actually **skipped**. Root cause was a pre-existing expression bug on the checkout: `fetch-depth: ${{ cond && 0 || 1 }}` always evaluates to `1`, because the number `0` is [falsy in GitHub expressions](https://docs.github.com/en/actions/reference/workflows-and-actions/expressions) — so `cond && 0` short-circuits and `|| 1` always wins. Both database legs had been checked out shallow all along. Fixed by quoting the depths, plus a missing PR base commit is now a **hard error** rather than a skip, so a shallow checkout can never again produce a vacuous green.

**Proof the guard actually fires:** a temporary commit editing a released changelog was pushed to this branch; the `Migrations` gate went red with the intended message, and the commit was then dropped from history:

```
failure  Changelog immutability guard
::error:: Released Liquibase changesets were modified, renamed, or deleted.
          Ship a new follow-up changelog instead: 
          server/src/main/resources/db/changelog/1731511193057_changelog.xml
```

On the current (clean) head the guard **executes** and passes for the right reason — `fetch-depth: 0`, `Baseline: 4692c8e3…`, `No released changeset was touched; master.xml stays append-only.`

---

**Out of scope, worth a follow-up:** `.github/workflows/ci-security-scan.yml:36` has the identical bug — `matrix.scan-type == 'secrets' && 0 || 1` — so the secrets scan has been running against only the latest commit instead of full history. Deliberately not bundled here: switching it to full history may surface historical hits and should land on its own.

Issue #1358, opened 2026-07-16
feat(server,webapp): delivery funnel and shadow feedback preview


FelixTJDietrich:
**Execution priority after #1423 (2026-08-03):** the full analytics funnel does not block a silent deployment. The pilot-critical slice is smaller: unmistakable not-delivered framing, the effective resolver reason, and the last-N exact feedback preview used by #1357's activation review. Build that slice with #1357; complete aggregate funnel analytics before broad rollout.

## Current state after #1423 (2026-07-31)

#1423 delivered the read foundation for Slice 2: workspace admins can filter review findings and exact stored feedback bodies by practice/person/channel/date/delivery state/suppression reason, open shareable details, and distinguish Delivered, Not delivered, and No feedback composed. The admin reads are tenant-scoped and permission-gated, but explicitly **not read-audited**.

Still open:

- Slice 1 funnel counts and click-through stages.
- The #1356 persisted evaluation/reason chain (today only the terminal suppression reason exists).
- Explicit, screenshot-safe **not-delivered** framing that names *why* — awaiting approval, suppressed by policy, backfilled, or failed. #1423's generic Not delivered treatment collapses all four. There is no per-practice `Shadow` state; the tiers are Off / Propose / Deliver.
- Backfill/replay provenance markers.
- Auditing or a recorded decision for reads of named people's AI assessments.

## Problem

Silent mode is only useful if admins can understand what the system *would have posted* and why. #1423 made individual stored findings and feedback readable, but there is still no funnel, persisted policy trace, or unmistakable not-delivered rendering. Observing reliability before rollout remains the point of the silent phase.

## What

One surface, two slices (per-practice analytics is the third slice, tracked in #1339):

- **Slice 1 — funnel counts:** detected → prepared → delivered / suppressed (by reason), grouped by practice and channel, time-windowed. Implemented as aggregation over the feedback ledger.
- **Slice 2 — undelivered preview:** filterable list of undelivered feedback rendered as it would have posted (composed comment body), per practice / user / channel, showing the resolver's reason chain.

Backfilled results (#1376) appear here alongside live undelivered results, clearly marked.

## Design references (researched patterns — guidance, not binding spec; invariants live in #1356/#1377)

[...]

Issue #1183, opened 2026-05-17
feat(server): per-role Grafana dashboards for app, ingest, worker, sandbox


FelixTJDietrich:
Part of #1114.

## What ships

Four Grafana dashboards JSON-provisioned under `infra/grafana/dashboards/`, one per deployment role.

Concretely:

- `infra/grafana/dashboards/role-app.json`: HTTP request rate, p95 latency, DB pool saturation, NATS publish rate, WSS connected workers, active mentor sessions.
- `infra/grafana/dashboards/role-ingest.json`: webhook rate by source, HMAC failure rate, dedup hit rate, NATS publish lag.
- `infra/grafana/dashboards/role-worker.json`: WSS reconnect rate, in-flight jobs, sandbox spawn p95, LRU evictions, capacity headroom, drain events.
- `infra/grafana/dashboards/role-sandbox.json` (host-level): container count, OOM rate, IMDS-block hit rate, Squid 403 rate.
- All four provisioned via the Grafana file-provider sidecar (no UI edits).
- Local docker-compose smoke loads all four in Grafana 11+ without errors and renders panels with synthetic data.

## Why

Without role-specific dashboards, the on-call has to mentally filter a single mixed dashboard by `service.role` for every query, every time. The per-role split is the first-mile usability fix that makes the metrics substrate from #1181 / #1182 actually load-bearing for incident response.

## Acceptance criteria

- [ ] `role-app.json`: panels = HTTP request rate, p95 latency, DB pool saturation, NATS publish rate, WSS connected workers, active mentor sessions
- [ ] `role-ingest.json`: panels = webhook rate by source, HMAC failure rate, dedup hit rate, NATS publish lag
- [ ] `role-worker.json`: panels = WSS reconnect rate, in-flight jobs, sandbox spawn p95, LRU evictions, capacity headroom, drain events
- [ ] `role-sandbox.json`: panels = container count, OOM rate, IMDS-block hit rate, Squid 403 rate
- [ ] All four provisioned via the Grafana file-provider sidecar — no UI edits required, no manual import
- [ ] Local docker-compose smoke test loads all four dashboards in Grafana 11+ without errors and renders panels with synthetic data

## Implementation notes

- File-provider provisioning is the rule, not the exception. UI edits drift silently; PR diffs against committed JSON don't.
[...]

PR #1423, opened 2026-07-28
feat(workspace): improve practice reviews and catalog administration

### LINKED ISSUE #1420: feat(server,webapp): admin read surface for observations and prepared feedback
## Problem

There is no admin-readable surface for what detection produced. `ObservationController` is caller-scoped (a developer sees only their own observations); there is **no REST endpoint over the feedback ledger at all** — every `FeedbackSuppressionReason` row is unreachable; and the agent-activity page renders job status, model and token counts but no findings, no composed body, and no suppression reasons.

Consequence: today the only way to inspect detection output is to read the comment it posted. The moment delivery is suppressed — by silent mode (#1355), by a shadow practice, or by any policy — the operator's visibility drops to zero. A silent phase whose output nobody can read proves nothing while still spending money.

## What

An admin read surface over data that already exists (no new detection concepts required):

- List and filter observations across the workspace (not just the caller's own), with their practice, assessment, severity, evidence and the job that produced them.
- For a given run or artifact: the prepared feedback, its delivery state, and its suppression reason — read straight from the existing ledger.
- Render the composed body as it would have posted, marked unmistakably as not-delivered where that is the case.

## Relationship to #1358

This is the unblocked half. #1358's per-reason funnel with click-through needs the full evaluated decision chain from #1356 and stays blocked on it; **this issue needs only the ledger that already ships**, which is why it can land before the silent deploy rather than after it.

## Dependencies

None. Extends the existing agent-activity admin surface — sequence with #1387, which touches the same page.

Part of #1351.

FelixTJDietrich:
## Description

Workspace admins can see that a practice review ran, but withheld or failed feedback is otherwise invisible. This PR adds a readable, admin-only review history and turns the practice catalog into a responsive administration surface. It also makes rollout safer: admins can prepare the catalog, model, and review rules while new reviews are paused, then start project reviews from the Review settings page.

The experience is organized around three operator questions:

| Section | Question it answers |
| --- | --- |
| **Review activity** | What ran, against which work, and what did it produce? |
| **Findings** | What strengths or improvements did the review identify, and what evidence supports them? |
| **Delivery** | What message was prepared, did it reach a surface, and if not, why not? |

It also lets admins reorder practices or move them freely between any practice area and **Unassigned**, including empty destinations, without losing changes on restart. Catalog placement is optimistic: rows and handles stay mounted at a stable size, the drag preview preserves the source row's dimensions, and consecutive moves are serialized without freezing unrelated controls. While an area is being deleted it disappears only as a destination; other safe moves remain available. Detail views use shareable routes, preserve full important text, and progressively reveal evidence, delivery metadata, and technical provenance.

Closes #1420.

### Implementation notes

- Adds workspace-admin, resource-oriented reads for review summaries, findings, and feedback details. Run and reviewed-work drill-down reuse the same filtered collections.
- Keeps Catalog, Practice feedback, Review settings, model configuration, and historical output available while new reviews are paused. The project-review switch lives with review triggers and rules rather than the generic workspace-capabilities card; activation remains unavailable until the review model is runnable.
- Keeps admin navigation stable for configurable and historical surfaces, gates only genuinely unavailable member destinations, and applies the same Heph eligibility predicate to its sidebar link, workspace route, and global launcher. Every member can reach their own practice-feedback preference.
- Treats feedback delivery as a ledger: **Delivered**, **Not delivered**, and **No feedback composed** remain distinct. “Delivered” means placed on a surface, not proven read.
- Uses routed navigation links with `aria-current` for **Review activity / Findings / Delivery**. These are separate documents with shareable URLs, so an ARIA tab widget would be the wrong interaction model.
- Defines page surface behavior at the route boundary: standard pages share one `px-4 py-6` content gutter, full-screen tools own the remaining flex space, and landing/auth pages keep their intentional shells. This removes nested page padding and fixed-header-height calculations without flattening every page into one layout.
- Uses the repository's shadcn/Base UI primitives for cards, items, accordions, menus, dialogs, popovers, selects, fields, switches, and responsive list states. Mobile catalog filtering becomes a Select; desktop keeps the compact Toggle Group. The area visual picker uses the existing Toggle Group composite for roving keyboard focus instead of a hand-rolled button grid.
- Uses the generated Hey API mutation and query-key helpers with TanStack Query's optimistic-mutation lifecycle: cancel affected reads, snapshot only placement fields, update list and detail caches, roll back on failure, reconcile from the server response, and invalidate after the last related write. Placement writes share a mutation scope so fast consecutive drops execute in order without overwriting one another. Pointer hit-testing prioritizes the practice row under the pointer, then its bucket, while area headers and practice buckets use separate droppable geometry; blocked destinations are removed from collision detection rather than accepted and rejected after the fact. A single idempotent `PUT /placement` atomically expresses the destination area and exact position; omitting `areaSlug` means **Unassigned**.
- Installs the default catalog exactly once per workspace. Later moves, deletions, and unassignments are not recreated by startup seeding.
- Audits practice-definition create/update/delete events with deterministic snapshots and SHA-256 fingerprints; raw criteria and precompute scripts are never written to the audit log.
- Consolidates every schema change into one branch Liquibase changelog and regenerates the ERD, OpenAPI specification, and TypeScript client.

## How to test

As a workspace admin:

1. Open **Practices → Reviews**. Move between **Review activity**, **Findings**, and **Delivery**; exactly one section should have the selected treatment and `aria-current="page"`.
2. Open a completed review. Confirm the reviewed work, findings, and delivery outcomes are understandable before expanding technical details.
3. Filter **Findings**, open a result, and verify its evidence, practice, developer, producing review, and linked feedback.
4. Filter **Delivery** to not-delivered messages. Open one and confirm the composed body remains readable inside the persistent **Not delivered** treatment with its reason.
5. Open **Practices → Catalog**. Throttle the network, then reorder within an area and move practices into another area, an empty area, and **Unassigned**. The source row and handle should stay mounted, the overlay should keep the row's width, and the exact drop position should update immediately. Make a second move before the first response returns, then refresh and confirm the final placements persist. While deleting an area, confirm it is no longer offered as a destination.
6. At 320 px, confirm the catalog uses a single filter Select, cards reflow without horizontal scrolling, names wrap instead of truncating, and Review settings have one consistent page gutter.
7. Turn project reviews off. Confirm **Catalog**, **Practice feedback**, **Review settings**, and the practice-feedback model remain available; no configuration action should silently start reviews. Confirm the status switch cannot be turned on until its model is runnable.
8. Sign in as a non-admin workspace member. Admin routes should redirect away and API requests should return 403. Confirm the member can still manage their personal practice-feedback preference.

## Verification

| Gate | Result |
| --- | --- |
| Formatting, lint, TypeScript, PMD | `pnpm run format && pnpm run check` passed |
| Webapp unit suite | 746/746 passed |
| Storybook browser suite | 1090/1090 passed, including strict accessibility, 320 px reflow, long-content, empty destinations, between-row placement, blocked targets, and real pointer drags |
| Architecture + Modulith | 224/224 passed |
| Fresh database migration + schema diff | 867/867 changesets applied; no Hibernate/Liquibase drift |
| Practice catalog integration | 79/79 passed, including cross-area, empty-destination, unassigned, authorization, and atomic-failure cases |
| Review-output integration | 39/39 invocations passed |
| Server unit tier | 5785/5785 passed |
| Production builds | Webapp and Storybook passed |
| Deployed 320 px overflow smoke | 24/24 mobile stories fit the viewport; app, Storybook, and Storybook index return 200 |
| Manual drag geometry | Source and overlay matched exactly at 978 px desktop and 270 px mobile with no horizontal overflow; the empty-area drop resolved to exact position 0 |
| Live smoke | App/API healthy; 75 monitored repositories and no duplicate SCM accounts. The current seed has no mentor-ready member, so a signed-in Heph smoke remains unavailable. |

## Checklist

- [x] No operator action is required; migrations and default-catalog installation are automatic

## Screenshots

- [Live test deployment](https://hephaestus-test.felixdietrich.com/)
- [Current Storybook](https://hephaestus-test.felixdietrich.com/storybook/)
- [Review navigation](https://hephaestus-test.felixdietrich.com/storybook/?path=/story/admin-practice-reviews-navigation--review-activity)
- [Review activity](https://hephaestus-test.felixdietrich.com/storybook/?path=/story/admin-practice-reviews-review-activity--default)
- [Not-delivered message](https://hephaestus-test.felixdietrich.com/storybook/?path=/story/admin-practice-reviews-message-details--not-delivered)
- [Responsive practice catalog](https://hephaestus-test.felixdietrich.com/storybook/?path=/story/admin-practices-catalog--populated)
- [Review settings](https://hephaestus-test.felixdietrich.com/storybook/?path=/story/admin-practices-review-settings--configured)
- [Area visual picker](https://hephaestus-test.felixdietrich.com/storybook/?path=/story/admin-practices-building-blocks-area-visual-picker--default)

## Limits and reviewer focus

- **Visual review:** the [current published Storybook](https://66a8981a27ced8fef3190d41-clvnolkwwt.chromatic.com/) was built from commit `2988507a5`. Chromatic passed; visual differences remain reviewable and were not accepted automatically.
- **Withheld inline notes:** no placement row exists when nothing was posted. The UI shows the finding's evidence anchors but does not invent an exact placement.
- **Read auditing:** these admin GETs are permission-gated but not audit-logged. This PR does not claim the audited-read machinery discussed in #1358.
- **Criteria provenance:** finding details expose the criteria revision selected as of job start. A criteria edit between job start and sandbox materialisation can make that revision differ from the text the detector received; this inherited race remains explicit rather than being presented as exact replay provenance.
- **Live authentication:** the public deployment intentionally disables passwordless dev login. The app and real SCM seed are live; use the public Storybook for deterministic UI review unless you have valid OAuth access.
- Suggested first review: `PracticeReviewsLayout.tsx` (route semantics), `PracticeCatalog.tsx` (responsive catalog), `FeedbackMessage.tsx` (untrusted markdown and delivery framing), `ObservationRepository` (workspace-scoped ranked disposition), and `PracticeDefinitionSnapshot` (audit redaction).


FelixTJDietrich:
## Review — verdict: merge after small fixes, no blockers

The security half is right: tenancy is enforced in SQL with the workspace as a **parameter** (never a filter), every handler is `@RequireAtLeastWorkspaceAdmin`, cross-workspace detail reads 404 rather than 403 (no existence oracle), and three ArchUnit rules pin the properties that would otherwise rot. No N+1 — identity resolution batches per page, bodies truncate in SQL (`left(f.body, 320)`), pagination clamps rather than rejects. Tests are real Postgres + real HTTP with zero mock assertions.

Standouts beyond the issue's ask: bidirectional observation↔unit drill-down; the per-row `LATERAL` disposition column; `ComposedBody` treating the body as untrusted LLM prose over attacker-authorable PR content (images dropped, non-`http(s)` hrefs demoted, raw HTML demoted — with a `play` test firing a real payload); `reachedAReader` as a total `Record` map so a new server-side state breaks compilation instead of inheriting a default; and the `openapi.jmx.port` fix, which repairs a silent build failure where a port collision produced a stale spec under `BUILD SUCCESS`.

### Should-fix before merge

1. **`artifactLabel` renders the internal primary key as a provider PR number** — `detection-format.ts:134-142` emits `Pull request #${artifactId}`, but `artifact_id` is the surrogate id (`PracticeDetectionDeliveryService:299-312`), not `Issue.number`. PR #42 with internal id 1183 displays as "Pull request #1183"; the operator searches GitHub and finds the wrong PR. On a surface built for cross-checking, a plausible-but-wrong identifier is worse than a raw id. Render `(id 1183)` or resolve `number` + `nameWithOwner` server-side.
2. **"there was nothing to say" is false for any run that didn't complete** — `RunDetectionSections.tsx:113-117` asserts unconditionally, while its sibling twelve lines up correctly hedges with *"if it completed"*. A run refused at claim time with `BUDGET_EXHAUSTED` has no observations and no ledger rows, so the operator reads a positive claim that nothing was withheld. Gate both empty states on `COMPLETED`. (`cancellationReason` is also absent from the API, so the panel can't name the reason — fuller fix, fair as a follow-up.)
3. **The changeset advertises two filters the UI cannot reach** — `date` (`from`/`to` exist server-side but are absent from `detectionSearchSchema`) and `person` (in the schema, but no control produces it). Changesets become the release note; drop the words or wire the controls.
4. **`CrossTenantIsolationIntegrationTest` not extended** — its own javadoc says new workspace-scoped read paths add a case. Coverage here is genuinely equivalent, but that suite shares an `account_login` across two workspaces with an overlapping member, so it catches org-login-scoped leaks that the detection test's distinct-org fixture cannot. ~40 lines.

### Follow-ups (fine after merge)

5. The furthest-reaching disposition collapse ranks `SUPPRESSED` below `PREPARED`, so a `VOLUME_CAPPED` finding rerouted to the conversational lane shows as "Prepared" with a blank reason — the list's primary scan misses it. Worth a line in `evaluation-provenance.md` §5 next to the ranking rule.
6. `EvidenceBlock` falls back to raw JSON whenever `locations`/`snippets`/`references` are all empty — but `normalizeEvidence` *always* emits `{locations:[], snippets:[]}`, so every `PRESENT/GOOD` observation renders a JSON blob instead of the "no structured evidence" line. Also, `references` is documented but never emitted.
7. Neither list has an index serving its sort (`feedback` is indexed on `workspace_id` alone; `observation` has no tenant column at all). The PR documents this honestly as a known ceiling with both fixes named — flagging only so it isn't lost.
8. Storybook covers the four components that matter; `EvidenceBlock` (which has the interesting fallback branch) is missing.

Issue #1374, opened 2026-07-16
refactor: remove leaderboard, leagues, and XP progression


FelixTJDietrich:
## Problem

The leaderboard/league/XP subsystem was superseded by the practice-centric model and is flag-gated off, but its code, schema, scheduled tasks (`LeaguePointsUpdateTask`, `SlackWeeklyLeaderboardTask`), the Slack digest publisher, and the `workspace.leaderboard*/leagues/progression` flags all remain. It is dead weight, an unnecessary tenancy/egress surface, and confusing to contributors.

## What

- Remove leaderboard/league/XP code paths (server + webapp), scheduled tasks, and the Slack digest publisher.
- Remove the related workspace feature flags and columns (including the `leaderboardNotificationEnabled` residue).
- Liquibase migration dropping tables/columns after a data snapshot; ERD and OpenAPI regenerated.
- The `leaderboard-flag-phase-out` branch (PR #1346) is the prepared precursor — it enforces the workspace flag and ports the read model dark; build the removal on top of it.

## Sequencing

Today `/w/{slug}` *is* the leaderboard page. The replacement developer-facing practice home is **Valentin's deliverable** (PR #1347 serves as reference, not a merge candidate) and gates Gate 2 — the code removal itself does not wait on it. The existing production instance still shows the leaderboard to real users: coordinate this removal's deploy so they are not left with an empty product in the interim.

## Concrete removal scope (from the page audit)

- `LeaderboardPage` route (`/w/$slug/index`), leaderboard components, league progress/scoring dialogs.
- `ProfileHeader` gamification (level badge, XP bar, league points) — and note `ProfileContent` **imports from `components/leaderboard/`** (`ActivityBadges`, `ReviewsPopover`, `score > 0` filter): extract/refactor these before deleting the directory (#1273 is the replacement surface).
- `AdminSettingsPage`: strip the Leaderboard/XP/Leagues feature toggles and the `AdminLeagueSettings` league-reset action (embedded remnants, not standalone pages).
- Scheduled tasks + Slack digest publisher (server side, as above); delete the `ai/agents` / `ai/practice-detection` redirect shims while touching navigation.

## Scope decisions

- **Achievements stay** (flag-gated): the `achievement/` package is self-contained — no XP/league/leaderboard coupling. Post-1.0 intent: re-evaluate achievements against the practice-centric model; either integrate or remove in v2.0.0 — this sentence is the named owner of that decision.
[...]

Issue #1368, opened 2026-07-16
feat(server,webapp): per-workspace LLM cost rollup and monthly budget cap


FelixTJDietrich:
## Problem

No per-workspace visibility of LLM spend and no budget enforcement — a single misbehaving workspace or sync burst can quietly burn the instance budget.

## What

- Per-workspace spend rollup (day/month, by job type): instance admin sees all workspaces; workspace admins see their own.
- **Monthly budget cap per workspace:** when exceeded, pause detection for that workspace and alert. Eventually-consistent checking is fine — a small overshoot is acceptable.

## Unified accounting (decided — pre-1.0 breaking changes welcome)

- Consolidate LLM cost into **one usage ledger** covering detection, replay/backfill, AND mentor turns (today mentor cost hides in `chat_message.metadata` JSON) — prefer the clean table + migration over querying JSON metadata forever.
- Rollup and monthly budget-cap enforcement both read that ledger; budget pause halts detection, replay/backfill, and (documented mechanism) mentor turns for the workspace.
- Per-user mentor turn quota remains v2; the workspace budget is the 1.0 backstop.

## Non-goals

Hard reservation/quota semantics, billing/chargeback, token-level rate limiting.

Part of #1354.

PR #1346, opened 2026-07-12
feat(leaderboard): enforce the workspace flag and ship the practice report read model dark

FelixTJDietrich:
## Description

Transition step of the leaderboard retirement (ADR 0023, phased). **No schema and no leaderboard code is deleted or renamed** in this PR; old app images keep working against the migrated schema. One practice API endpoint is replaced (details below).

**1. The legacy leaderboard is now actually flag-gated.** The existing per-workspace `leaderboardEnabled` / `leaguesEnabled` flags become the server-side enforcement point: flag-off workspaces get 404 `ProblemDetail` from the leaderboard/league endpoints (including the admin league reset), and the weekly job (Slack digest + league-points recompute) skips them. New workspaces default both flags off (already the stored default); existing workspaces keep their current values — nothing changes for them on upgrade. The admin toggle copy now marks the leaderboard as a legacy feature being retired in favor of practice feedback.

**2. The practice report read model ships dark** (endpoints, schema, generated client; no navigation, no UI):

- `GET /practices/reports/me` — the developer's own per-practice cards (status + cycle-over-cycle trend), server-gated to the caller
- `GET /practices/reports` + `/practices/reports/{userId}` — admin/owner-only roster and drill-down; every served response writes an append-only `data_access_event` disclosure row (actor, subject, resource, timestamp; storage-layer append-only trigger; purged with the workspace, redacted on account erasure)
- `GET /practices/health` — k-anonymised (K=5, compound small buckets) workspace health per practice area; member access gated by the new `healthVisibility` setting (MENTORS_ONLY default); admin reads unsuppressed (they already see the named roster)
- `NonCompetitiveSurfaceArchTest` pins that no practice DTO field is named like a score/rank/ELO/XP/league
- The review-cycle resolver reads the existing `leaderboard_schedule_day/time` columns — the rename belongs to the final retirement PR
- JIT OAuth account creation binds `identity_link.external_actor_id` eagerly via a new `ExternalActorQuery` SPI (exact case-insensitive match, unique-match-only — an ambiguous login binds nothing and never breaks the login)

**API replacement (breaking for external clients only):** `GET /practices/observations/reflection` is removed in favor of `/practices/reports/me` (same cards, plus trend). The bundled webapp never consumed the old endpoint.

**Phased plan (ADR 0023):** flag-gated retirement now → dark read model now → user-facing surfaces after a design pass → deletion of leaderboard code/columns as a later, separately-reviewed decision gate.

[...]

Issue #1340, opened 2026-06-30
feat(webapp): developer reflection dashboard (per-practice observations and feedback)


FelixTJDietrich:
## Problem

Developers receiving feedback have no in-app place to see, per practice, where they stand and what has been delivered to them. The reflection APIs (`/practices/observations/reflection`, `/summary`, detail) exist with zero webapp consumers — delivered feedback is a comment with no home. The mentoring-not-monitoring commitment (#1377) needs a developer-first view, not only admin analytics.

## Decided

- **Learner-safe rendering:** detection `criteria` are never shown to a learner; disclosure is progressive — practice name, why-it-matters, and feed-forward first; what-good-looks-like, area context, and trend on request.

## What

- A developer (non-admin) route with per-practice cards: why-it-matters, what-good-looks-like, presence×assessment trend, and the feedback delivered to this user — with inline reactions (ADDRESSED / DISPUTED / NOT_APPLICABLE).

## Non-goals

Admin aggregates (#1339); public profiles (#1275).

## Dependencies

APIs exist; coordinate with #1273 and the practice-overview surface family.

## Ownership

Valentin (developer-facing practice surfaces).

FelixTJDietrich:
Kept in v1.0.0 as a stretch item: the reflection API exists and is unconsumed, making this one of the cheapest real features on the list — but it yields to the rollout/eval/ops clusters if time is short.

FelixTJDietrich:
Capacity pass 2026-07-16: cut to v2.0.0. The milestone was ~1.9× oversubscribed against measured throughput; a 'stretch' item inside an oversubscribed milestone is denial with extra steps. The reflection APIs remain unconsumed and this stays the cheapest post-1.0 feature.

FelixTJDietrich:
Restored to v1.0.0 (mission recentering): the developer-facing reflection surface is core to the mentoring mission — a 1.0 that is all admin consoles and PR comments reads as monitoring, not mentoring. Sequencing: after PR #1347 merges (shared navigation), parallel to admin surfaces.

FelixTJDietrich:
Ownership note (2026-07-16): the developer-facing reflection dashboard is Valentin's task, together with the practice-overview surfaces (PR #1347 serves as reference, not a merge candidate). Stays in v1.0.0 — a developer-facing practice home gates Gate 2 (pilot users need somewhere to see their feedback context).

FelixTJDietrich:
[...]

PR #1424, opened 2026-07-28
feat(server): practice reports, and a record of who else has read yours

### LINKED ISSUE #1421: feat(server): non-competitive practice report read model with access disclosure
## Sequencing update after #1425 and the evaluation-model design (2026-08-02)

Hold PR #1424 until #1428 establishes the evidence/semantic/execution outcome model and #1429 names the persisted-observation lineage stage. This read model currently queries the legacy collapsed observation semantics; rebasing it first would create report contracts and health calculations that immediately need rewriting. After those contracts land, salvage the disclosure/access work and rebuild report denominators against the new model. It remains independent of #1374.

## Problem

The non-competitive practice report read model — per-user and workspace practice reports, a health endpoint, disclosure records for data access, and the architecture test that keeps these surfaces non-competitive — was built on the `leaderboard-flag-phase-out` branch and never landed. It is the data dependency for the developer-facing practice surfaces, and it is fully additive: nothing in it depends on the leaderboard that #1374 removes.

## What (salvage, do not rebase the original branch)

Cherry-pick onto a fresh branch off `main`:

- `/practices/reports/me`, `/practices/reports`, `/practices/reports/{userId}` read model and the practice health endpoint.
- Data-access disclosure records for report reads, with the anonymity floor applied to aggregate views.
- The architecture test asserting these surfaces stay non-competitive, and the external-actor query SPI it relies on.

Explicitly **not** salvaged: the leaderboard flag-gating half of that branch (returning 404 from controllers that #1374 deletes, and retirement copy on toggles that are already off by default for every workspace). It buys no operational safety and conflicts with code that is about to be deleted.

## Dependencies

Independent of #1374; sequence before or alongside it. Consumed by the developer-facing practice surfaces.

Part of #1377.

FelixTJDietrich:
## Description

Developers who receive AI practice feedback have had nowhere in the product to read it. This adds the read model behind that: **a practice report** — the same findings that arrive on pull requests and from the mentor, reorganised per practice so a person can sit with them.

Three surfaces, one derivation:

| Endpoint | Who | What |
|---|---|---|
| `GET /practices/reports/me` | any member | your own per-practice cards: why the practice matters, what good looks like, where you stand, what to act on, what you already do well |
| `GET /practices/reports` | admin / owner | roster of everyone with recent activity, people who may need support first |
| `GET /practices/reports/{userId}` | admin / owner | one developer's cards — *exactly* the ones they see |
| `GET /practices/health` | admin / owner, or members if opened | anonymised per-area distribution, names nobody |

**Nothing here ranks people against each other.** No score, no rank, no percentile, no total — only where someone stands against each practice's own standard, and how that moved since the previous window. That is the whole point of replacing the leaderboard rather than recolouring it, so it is written down in [ADR 0028](docs/decisions/0028-criterion-referenced-practice-reports.md) and pinned by an ArchUnit test that fails the build on a DTO field named like a score.

**Opening someone else's report is recorded, and they can read the record.** Every admin view of a named report writes an append-only row; the subject gets their own copy in the GDPR data export — when, in which workspace, and who did it. [GDPR Art. 15(1)(c)](https://gdpr-info.eu/art-15-gdpr/) entitles a data subject to know who received their data, and the CJEU held in C-154/21 that this means the recipient's *identity*, not just their category. There is deliberately **no admin-facing view** of that trail: "what has this admin been reading" is a surveillance question wearing a compliance hat.

Reading your own report records nothing.

<details>
<summary><b>Where this came from</b></summary>

The read model was built on the closed `leaderboard-flag-phase-out` branch (PR #1346) and never landed. This salvages it, re-designed where it had gone wrong:

- It anchored the report window to `workspace.leaderboard_schedule_day/_time` — the columns #1374 deletes, and the leaderboard's weekly reset leaking into the thing meant to replace it. Now a rolling 28-day window (`PRACTICE_REPORT_WINDOW_DAYS`), which also fixes the real product bug: a one-to-two-week window left ordinary contributors with an empty report.
- Its anonymity rule caught "almost none" but not "almost all" — 20 of 20 developers at the same status clears any group-size floor while disclosing every individual.
- The self-view resolved its subject by login, which on a publicly-readable workspace could serve one person's report to someone sharing their login on another provider.

</details>

## Review guide

`server/openapi.yaml` and `webapp/src/api/**` are regenerated, not hand-written — about 840 of the added lines. The API delta itself is small: four paths added (`/practices/reports`, `/practices/reports/me`, `/practices/reports/{userId}`, `/practices/health`), one removed (`/practices/observations/reflection`), and the report schemas. The rest of that diff is the generator re-serialising unchanged content.

The code worth reading, in order:

1. `practices/report/PracticeReportService.java` — the read model, including the anonymity rules
2. `core/audit/access/` — the disclosure trail
3. `db/changelog/1785211534351_changelog.xml` — new table, append-only trigger, index rebuild
4. `docs/decisions/0028-…md` — the reasoning, so the code does not have to carry it

## Notable decisions

**The migration rebuilds an index on `observation`, concurrently.** Widening `idx_observation_subject` to `(about_user_id, observed_at DESC)` with a plain `CREATE INDEX` takes an `ACCESS EXCLUSIVE` lock for the whole build — a write outage on the highest-volume table in the schema, which detection writes to continuously. It builds `CONCURRENTLY` in a `runInTransaction="false"` changeset, then drops the narrow index, so the table is never without one. Verified against a real Postgres: index present, zero invalid indexes, zero JPA-vs-DB drift.

**`GET /practices/observations/reflection` is replaced by `GET /practices/reports/me`.** Pre-1.0, and it had zero webapp consumers. One noun for one concept.

**A pre-existing bug is fixed on the way past.** `findRecentByDeveloperAndWorkspace` correlated its latest-run subquery on the artifact alone, so a co-author's later re-review became "the latest run" for everyone on that pull request — erasing a developer's findings from their own history *and* from the mentor's context. Now correlated on `(workspace, artifact, about_user)`.

**Server-only, deliberately.** `healthVisibility` is settable on `PATCH /workspaces/{slug}/features` but has no UI. I built an admin toggle for it and then removed it: `GET /practices/health` has no client, so the switch would have produced a success toast and no observable change. It lands with the screens that render it. The developer-facing surfaces are #1340 / #1273.

**I did not split `PracticeReportService` (534 lines, three surfaces).** They share the window, the status deriver and the area rollup; splitting them invites the anonymity and status rules to drift apart, which is the failure the file exists to prevent. Reasonable people may disagree — happy to split if you want it.

## Risks and follow-ups

- The disclosure trail has no in-app read surface beyond the data export. A person-facing "who has seen my report" page is #1196's scope.
- `LegacyAgentConfigMigrationIntegrationTest` assumed the #1368 changelog is `master.xml`'s tail, which breaks for *any* branch appending one. Fixed here; the same fix exists on the #1323 branch, so expect a small conflict whichever lands second.
- `LoginProviderAdminControllerIntegrationTest` flakes under full-suite load (5s blocking read). Passes in isolation, unrelated to this branch.

Closes #1421

## How to test

```bash
cd server
mvn -o test -P'!quick'                                    # unit
mvn -o test -P'!quick,architecture-tests'                 # ArchUnit + Modulith
MANAGEMENT_PORT=39082 SERVER_PORT=39080 mvn -o verify -P'!quick' \
  -Dspring-boot.start.skip=true -Dspring-boot.stop.skip=true -Dspringdoc.skip=true
cd .. && pnpm run check && pnpm run test:webapp
```

**By hand, as a workspace member:** open `/practices/reports/me`. You get one card per practice with a `status` and a `trend`; `SELECT count(*) FROM data_access_event` stays at zero.

**As an admin:** `GET /practices/reports` and `/practices/reports/{userId}`. Each adds exactly one row naming actor and subject. Then confirm the trail resists tampering — on a `prod`-context migrated database, `UPDATE data_access_event SET resource_type = …` and `DELETE FROM data_access_event` both raise `append-only`.

**Migration safety:** against a throwaway Postgres, `mvn -o liquibase:update -Dpostgres.port=<port>`, then check `SELECT count(*) FROM pg_index WHERE indisvalid = false` returns `0` and only `ix_observation_subject_observed_at` remains. `mvn -o liquibase:diff` must report `changeSets count: 0`.

**Anonymity:** with fewer than five active developers in an area, a member gets `SUPPRESSED` and an admin gets real counts.

<details>
<summary><b>Test results on this branch</b></summary>

| Tier | Result |
|---|---|
| Unit (`@Tag("unit")`) | 5784 passed |
| Architecture (ArchUnit + Modulith) | 225 passed |
| Integration (`@Tag("integration")`) | 1226 passed |
| Webapp (Vitest) | 690 passed |
| `pnpm run format && pnpm run check` | clean |
| `liquibase:diff` (JPA vs DB drift) | 0 changesets |

</details>

## Checklist

- [x] If the operator must act on this change (new required env var, manual migration step), the changeset summary says how (`**Operators:** …`) and `MIGRATION.md` is updated — *no action is required here: the migration applies automatically and `PRACTICE_REPORT_WINDOW_DAYS` is optional with a default, so `MIGRATION.md` is untouched. The changeset names the variable anyway.*


FelixTJDietrich:
## Review — verdict: merge after fixes, no blockers

I went hunting specifically for the three blocking-class risks and found none of them. **No composable ranking**: members reach exactly one named surface (`/reports/me`); roster and drill-down are admin-gated, and the drill-down 404s through `requireVisibleSubject` on the same visibility floor, so id-guessing can't reach a hidden member. Nothing comparable is on the wire — `PracticeReportSummaryDTO` carries a boolean plus prose, and `attentionCount` never leaves the server. **The k-floor is sound including the complement case** (I verified the ADR's "complement < K" rule and the code's `count == total` are equivalent given primary suppression holds; whole-card rather than per-cell suppression is also correct, since four cells plus a visible total make one suppressed cell derivable by subtraction). **Tenancy holds** on all four endpoints. All 46 checks pass including the empty→head Liquibase replay.

Genuinely better than asked: the purge adapter *closes* the `agent_job`/`llm_usage_event` mistake instead of repeating it (order 1000, SPI-reached, transaction-local carve-out); the append-only trigger is actually executed against a real Postgres with `contexts=dev,prod` — the only tier in this repo where a `context="prod"` changeset runs — and attacked eight ways including erasure-as-cover-for-rewrite; `CREATE INDEX CONCURRENTLY` build-then-drop so `observation` is never unindexed and never takes an ACCESS EXCLUSIVE lock; and two real bugs fixed in passing (co-author latest-run correlation, self-view resolving subject by login = cross-provider disclosure).

### Should-fix

1. **`healthVisibility` is settable on an `@Audited` endpoint but invisible in the audit record.** `WorkspaceAuditSnapshots.FeaturesSnapshot` still enumerates only the eight booleans, so a PATCH widening health visibility to `EVERYONE` produces a `WORKSPACE_FEATURES` row whose before/after are byte-identical. In a PR whose thesis is "reads of assessed work are accountable", the one change that widens who can see what is the one the config ledger can't reconstruct. Also untested. Add the field to the snapshot + one IT.
2. **The disclosure trail isn't write-complete.** `ObservationController.getObservationsForPullRequest` returns every developer's BAD/ABSENT findings on a PR to any member, unscoped, and writes no `data_access_event`. Meanwhile the export presents `dataDisclosures` as the Art. 15(1)(c) answer. Alice's export would say nobody saw her feedback when Bob did. Either record the disclosure there or narrow the export wording.
3. **Cross-provider login hop on erasure/export.** `AccountPurger:137-145` and `ExportBundleAssembler.dataDisclosures` resolve actors via `membershipsForLogins` (no provider scoping) — the exact hazard this PR closes in `CurrentAccountUsers.resolveMemberOf`. With GitHub-`alice` and GitLab-`alice` both members, the export can hand one another's disclosure rows, and the purge can NULL the wrong person's actor ids. Use the provider-scoped `ExternalActorQuery#findExternalActorId` this PR introduces.
4. **The k-boundary is untested — both threshold mutations survive.** Cases cover 4, 20-unanimous, 19+1, 6/6/6; flip either `<` to `<=` and all still pass. Add an exactly-5 bucket and a total-of-5 case.
5. **Purge test creates a second tenant's row and never asserts on it** (`DataAccessEventImmutabilityIntegrationTest:204` — `otherTenant` unreferenced). The property that makes the `SET LOCAL` carve-out safe (scoped hole, not global) is asserted nowhere.
6. **The co-author correlation fix ships without a regression test.** The `AND f2.about_user_id = f.about_user_id` addition to `LATEST_RUN_WITHIN_WINDOW` fixes a real bug the PR describes well, but no two-subjects-one-artifact scenario exists in the diff — and it's a shared SQL constant used by three queries.
7. **`CrossTenantIsolationIntegrationTest` not extended** — the highest-value case (workspace-B admin drilling into a workspace-A user id) is tested nowhere; `unknownSubjectIsNotFound` uses a same-workspace member, a different property.

### Nits

8. The report window is 28d while the endpoint it replaces used 90d (`ObservationService.REFLECTION_LOOKBACK_DAYS`), and the mentor's context window is unchanged — so a developer's report and their mentor can disagree about "recent". Defensible; worth stating in the PR body, which currently reads as if it only widens.
9. `NonCompetitiveSurfaceArchTest` is real (record components are fields, so `int score`/`rank`/`percentile` fail the build today) but scoped to `..report.dto..` while `PracticeStatus`/`PracticeTrend` live one package up and serialise; and `COMPETITIVE_SEGMENTS` has `"standings"` but not `"standing"` — which is this codebase's own word (`PracticeStandingRow`, `findPracticeStandingForDeveloperBetween`).
10. `ObservationRepository:645` javadoc claims repositories-hidden-from-contributions are filtered; the SQL has no such predicate.
11. `SET LOCAL hephaestus.audit_purge = 'on'` is never reset — safe only because the adapter runs last at order 1000; a future contributor at 1001 inherits an unopposed DELETE hole. One `'off'` after the DELETE makes it order-independent.
12. `NO_DATA` is distinguishable from `SUPPRESSED`, a one-bit existence leak the ADR doesn't name; and the daily-rolling window with no noise permits differencing two k-safe releases. Low risk (default `MENTORS_ONLY`, no client) — one sentence in Consequences.
13. `LATEST_RUN_WITHIN_WINDOW` is a per-row correlated subquery inside the `filtered` CTE, run twice per `listReports`; `about_user_id` isn't in the index it probes. Fine now, first thing to watch as volume grows.
14. Orphaned javadoc block above the new one in `LegacyAgentConfigMigrationIntegrationTest` (the fix itself is correct and *strengthens* the test — counting forward to the release changelog rather than backward from master.xml's tail).

FelixTJDietrich:
## Sequencing after #1423

This PR is still valuable, but it is no longer next. Marking it draft while the Gate-1 safety path lands.

Do not rebase yet. After the #1386 shell and #1355 gateway merge:

1. rebase once onto the then-current `main`;
2. resolve the real `ObservationRepository` / `ObservationService` / report-model conflicts deliberately;
3. apply the existing review should-fixes (audit visibility, disclosure completeness/provider scoping, k-boundaries, purge isolation, correlation and cross-tenant coverage);
4. retimestamp/consolidate the one branch changelog and regenerate ERD/OpenAPI/client.

#1423 already supplies the admin shadow-read surface; this report model is not required for the first self-only shadow run.

Issue #1356, opened 2026-07-16
feat(server,webapp): bounded practice-review rollout and explainable delivery policy


FelixTJDietrich:
## Goal

Complete the safe path from assisted review to a deliberately bounded automatic rollout:

```text
Off → Review before sending → Send automatically
```

The control model should answer three different operator questions without conflating them:

1. **Should Hephaestus spend compute reviewing this work?** — rollout scope.
2. **How much authority does this practice have?** — practice autonomy.
3. **May this already-prepared feedback leave Hephaestus now?** — delivery policy.

Part of #1351. Builds on #1442's fail-closed instance egress boundary and #1444/#1451's evidence-aware review model.

## Shipped in #1453

#1453 delivered the core assisted-review path:

- one inherited `PracticeAutonomy { OFF, HUMAN_APPROVAL, AUTOMATIC }` model across workspace → area → practice;
- `OFF` excludes the practice from detection;
- safe default `HUMAN_APPROVAL`;
- immutable exact proposals with OWNER/ADMIN approve/reject, actor/time, rejection metadata and content digest;
- concurrent-decision protection;
- delivery-time autonomy and consent re-checks;
- terminal suppression under Silent Mode or lost eligibility;
- prospective transitions—old proposals never become automatic after a setting change;
- provider-marker reconciliation and recovery for approved SCM feedback;
- removal of the old assignee/`runForAllUsers` switch.

Do not recreate `Shadow`, `isActive`, `PracticeDeliveryMode`, or delivery state on observations.

## Remaining problem

For an AET Development pilot, running every practice over every repository and every contributor is unnecessarily expensive and widens the privacy and operational blast radius. Conversely, a delivery-only user allowlist still pays to analyze work whose feedback cannot be used.

Hephaestus therefore needs a small **review population** control before broad automatic rollout:

- selected repositories/branches;
- selected people whose work is in scope;
- practice autonomy;
- a workspace delivery pause.

Repository and person scope should reduce work **before expensive evidence capture/model execution** where the subject is known, while delivery policy must still re-check identity, consent and current scope before any write. Scope is operator targeting; personal consent remains an independent veto.

### 1. Workspace delivery pause

Provide one workspace-level emergency/maintenance control that stops external practice feedback without stopping detection or deleting targeting configuration.

- Pause is easy and immediate.
- Resume is prospective and never flushes queued/suppressed historical work.
- Instance Silent Mode remains the stronger instance-wide brake.

### 2. Review population

[...]

PR #1443, opened 2026-08-03
feat(workspace): adopt catalog practices without leaving practice setup

### LINKED ISSUE #1362: feat(server,webapp): workspace adoption of curated practices with provenance
## Current implementation status (2026-08-03)

PR #1443 is open and CI-clean. Its catalog browse/preview/adopt flow, optimistic ETag handling, duplicate protection, audit coverage, provenance, and “Not yet validated” presentation are the right product slice.

**Hold the PR for coordinated rework before merge:**

- #1356 owns the one rollout state, shipped as `PracticeAutonomy {OFF, HUMAN_APPROVAL, AUTOMATIC}` — Off / Propose / Deliver. Do not introduce a parallel `is_active` + `delivery_mode` state model. There is no per-practice `SHADOW` state: `HUMAN_APPROVAL` composes real draft feedback and queues it for a human decision.
- Whether a unit is delivered is a delivery-policy decision, not an `Observation` semantic. Persist the resolver/policy trace rather than hiding undelivered observations behind repository-wide delivery filters.
- The current historical occurrence-key lookup can let a previously delivering occurrence make a new proposed occurrence eligible after a Deliver → Propose transition. Eligibility must be scoped to the current resolved event/ledger decision.
- Rebase after #1444 settles the practice evidence/fingerprint schema, and integrate with #1356 so adoption lands directly at `HUMAN_APPROVAL` (Propose).

This still does not block the first silent deployment: #1442 provides the instance brake and existing workspaces/catalog snapshots can produce operator-visible shadow data. It is required before onboarding more workspaces or beginning the first controlled pilot.

## Problem

Workspaces should adopt catalog practices deliberately. Workspace administrators also need to know whether a practice came from the instance catalog and whether its review rules still match the source they evaluated.

## What

- Browse the instance catalog from workspace administration.
- **Adopt** copies an included catalog practice into the workspace.
- Record provenance as the catalog practice's immutable slug plus its review-rule fingerprint. Do not add a foreign key to catalog state: the effective catalog is computed from repository defaults and sparse instance overrides, and an adopted workspace copy must survive later exclusion or removal.
- Adopted practices behave like any workspace practice: they remain editable and start at Propose (`HUMAN_APPROVAL`).
- Show reliability metadata for the exact review-rule fingerprint at adoption time, or “Not yet validated” when none exists.
- When local review-rule edits change the fingerprint, replace the reliability numbers with **Changed since validation**. Guidance-only edits do not invalidate reliability metadata.
- Preserve the existing workspace labels from #1361: ordinary matching copies have no badge; exceptions use **Customized for this workspace**, **Instance catalog changed**, or **No longer included in new workspaces**.

## Acceptance criteria

- Adoption is explicit; catalog changes never rewrite workspace practices.
- The API and UI show the full incoming definition before adoption.
- Duplicate adoption is prevented by workspace slug identity.
- Provenance remains readable when the source is later customized, excluded, or removed.
- Accessibility does not depend on drag gestures, hover, or color.
- Configuration-audit rows cover adoption and resulting workspace configuration changes.

## Non-goals

Automatic re-sync or one-click replacement when the instance catalog changes.

## Dependencies

Depends on #1361 (catalog). Adoption should offer the enablement backfill (#1376) as the natural next step.

Part of #1352.

FelixTJDietrich:
## Description

Workspace administrators choose which practices their workspace reviews.

Until now a workspace received a copy of the whole instance catalog and there was no way to add a practice you had passed over. **Practice setup** now shows the catalog beside the practices you already have: read a practice, add it, and what you get is an independent copy that later catalog changes never rewrite.

Nothing sends feedback because you added it. A practice Hephaestus can review starts at **Review before sending**; one it cannot review stays **Off** until you connect what it reads.

**Reading no longer means editing.** Selecting a practice — yours, or one the catalog offers — opens it in a panel beside the tree you were reading. Selecting something inside that panel stacks a second one. Escape, a press on the page, the browser's Back button or a swipe steps back one level, and every panel is in the URL, so the view you are looking at is the view you can share.

### What a practice says, and in what order

`criteria` is the review rule, and it is written to the model that applies it: **35 of the 37** bundled practices contain `MAJOR` and `PRESENT`, 28 discuss severity, and the server composes a work-type preamble into it — so what reaches the screen has a median of **8,722 characters**. It was the headline of the adoption panel, at `text-xl font-semibold`, above the two fields written for a human.

Now `whyItMatters` leads, `whatGoodLooksLike` follows, and the rule sits under **How it decides** beside the precompute script. It is markdown — the editor has always promised that — and it renders instead of arriving as literal `##` and `-`.

**"Not independently validated" is gone.** `PracticeAutomatedReviewValidationStatus` has exactly one value, so the badge printed identically on all 37 practices while reading like a warning. In its place is a sentence saying nobody has measured how often the practice is right, and that this is why it starts by asking you to approve each piece of feedback.

**Provenance says what actually happens.** A copy never tracks the catalog, edited or not — `UPDATE_AVAILABLE` is computed for display and nothing in the product applies it. So the badges are **Same as the catalog**, **Catalog changed, yours did not**, **No longer in the catalog** and **Edited here**, and the matching case is labelled rather than silent.

### Behaviour changes

| Change | What it means |
|---|---|
| Automatic installation becomes repair-only | Existing workspaces keep everything. Any workspace with no recorded installation is still seeded once at startup — see the open question below. |
| Removing a group asks what to do with its practices | Keep them in Unassigned, or delete them with the group, which also deletes their observations. Previously they were always kept. |
| `…/practices/available` and `…/available/{slug}` retired | Both redirect into Practice setup; the second reopens that practice's panel. Existing links keep working. |
| One word throughout: **catalog** | One component had called itself "Practice catalog", "Practice library" and "the catalog" in three places. The repo's vocabulary doc already settled it. |
| The user-facing noun "area" becomes **group** | It was doing no work: three separate places explained an area *as* a group. `area`, `areaSlug`, `PracticeArea` and the API are unchanged. |
| Groups take an icon and a colour | The picker the instance catalog already had. Left alone, both follow the name. |

### API

Five operations under `/workspaces/{workspaceSlug}/practice-catalog/adoption`: list what is available, preview a practice or a group, adopt a practice or a group. Both adopt operations require `If-Match` — `428` without one, `412` when the validator no longer matches — so the UI can only apply a plan the administrator saw. A `*` wildcard is accepted as RFC 9110 requires; the first-party flows always send the preview's validator. Both write a configuration-audit row.

Provenance is the catalog slug plus the review-rule fingerprint, with no foreign key into catalog state: the effective catalog is computed from repository defaults plus sparse overrides, so a copy must stay readable after its source is customised, excluded or removed.

### Interaction, loading and reflow

- **Scroll.** The router resets scroll on every commit, including a search-param-only one, so opening a panel from row 40 threw you to the top and dismissing restored the old offset *after* the exit animation. Both navigators opt out, and `DetailStackLink` omits the prop so a future surface cannot reintroduce it.
- **The nested-panel jump.** `height` was in the popup's transition list, and Base UI releases a parent's pinned height on the child's completion frame. Height is now scoped to the axis where a drawer's height can legitimately change.
- **Motion** is 280ms in / 200ms out on split curves, down from a near-symmetric 450/400. Reduced motion drops the scale and the travel and keeps the fade.
- **Loading** is skeletons that hold the shape of what is coming. `LoadingBlock` is deleted — its `min-h-32`/`min-h-64` reserved a height matching nothing, which is the jump a skeleton exists to prevent.
- **Reflow.** `Field` ships a `responsive` orientation; the practice surfaces used hard `horizontal`, so a 14rem `Select` sat beside its label at 320px. Wide controls stack; switches and radios do not.

### Reviewer's map

172 files. Most of it is in five places:

| Where | What |
|---|---|
| `server/…/practices/curated/adoption/**` | The adoption API. Start at `CatalogAdoptionService`; `CatalogAdoptionPlanAssembler` turns a preview into a plan. |
| `server/…/PracticeCatalogInstallationManager.java` | Renamed from `DefaultPracticeCatalogSeeder`. Records the installation instead of filling the workspace. The repair path is unchanged. |
| `webapp/src/components/core/detail-drawer/**` | The stacking-panel primitive, its URL encoding, and the open/close hook. Nothing in it is practice-specific. |
| `webapp/src/routes/…/admin/practices/index.tsx` | Where the tree, the catalog and the panel stack meet. |
| `docs/admin/practice-catalog.mdx` | New. Adoption had no page for the people who do it — `grep -i adopt` over `docs/` hit one contributor file and nothing in `admin/` or `user/`. |

`webapp/AGENTS.md` gains the rules this work encodes: drawer-or-route, the loading ladder, motion tokens, search params that are UI state, the vocabulary table, status registries, and field orientation. `webapp/src/api/**`, `routeTree.gen.ts` and `server/openapi.yaml` are generated — skip them.

### Open question, found by deploying this branch

Against an empty database a fresh instance **still receives the whole bundled catalog in every workspace** (12 groups, 37 practices each). That is not what the adoption design intends.

Only `WorkspaceService.createWorkspaceWithInitialization` publishes `WorkspaceCreatedEvent`, which is what records the installation; `WorkspaceProvisioningService` and `GithubLifecycleListener` call `createWorkspace` directly, so workspaces provisioned at boot have no record and the repair pass seeds them. So the rule today is: **at startup, any workspace with no installation record gets the full catalog.** The docs say exactly that. Whether a fresh instance *should* come up empty is a product decision, not one to settle inside this PR.

### Not in this PR

- **No automatic re-sync**, and no one-click replacement when the catalog changes — an explicit non-goal of #1362.
- **No reliability numbers.** #1362 also asks for a *Changed since validation* state; there is nothing to invalidate until measured records exist. Stable catalog identities and technology variants are tracked in [#1445](https://github.com/ls1intum/Hephaestus/issues/1445).
- **A refused form submit announces twice** — the error summary and the inline field error are both live regions. Fixing it means changing the shared `ui/field` primitive.
- **The spacing scale is a rule, not a sweep.** 31 distinct values express about six relationships; the scale is written down, and converting every surface would repaint the whole diff.
- **Comment density is 12.2% of added lines against a 6.9% house baseline.** This round cut two dead modules, sixteen vocabulary violations, a self-contradicting pair of comments, ten restating labels and one sentence duplicated across twelve docs pages. What remains is mostly Base UI and router internals that are not derivable from the code; I stopped rather than delete load-bearing traps to reach a number.

Fixes #1362

## How to test

1. **Workspace administration → Practice setup → Show catalog.** Available practices appear grouped, each row carrying its own reason; ones you already have are not offered twice.
2. Select a catalog practice. The panel leads with why the habit matters; the rule is one disclosure away. Add it and you land back in the catalog, not in an edit form.
3. Select one of your own practices — read-only, with editing a separate marked step. Choose **Edit practice**, save, and confirm you return to that panel rather than a bare list.
4. Open a second panel from inside the first. Step back with Escape, a press on the page, Back, and a swipe. Each removes one level and the panel slides out.
5. Scroll halfway down the list before opening a panel. The page must not move, on open or on dismiss.
6. Add a whole group: every add, reuse and skip is listed before you apply it, and a failure leaves nothing partly created.
7. Change the catalog in another tab between opening a preview and applying it. Both the practice and the group flow refuse the stale plan and refresh in place.
8. Remove a group and choose **Keep practices unassigned**; then remove one and choose **Delete group and practices**.
9. At 320px: rows that pair a label with a dropdown stack instead of squeezing, and nothing needs a sideways drag.
10. Set your OS to reduced motion and repeat step 4 — panels fade, nothing scales or travels.

Automated: 12 server integration tests over the adoption controller (authorization, tenant isolation, complete previews, atomic group adoption, provenance, re-adoption, audit rows, missing and stale preconditions, duplicates, concurrency), 7 webapp route tests, and the Storybook suite below.

## Checklist

- [x] If the operator must act on this change (new required env var, manual migration step), the changeset summary says how (`**Operators:** …`) and `MIGRATION.md` is updated — no operator action is required; existing workspaces are untouched

## Screenshots

Chromatic sign-in required. These point at one build; the **Preview / Storybook** check always links the newest.

**The flow**
1. [Practice setup with the catalog shown](https://66a8981a27ced8fef3190d41-orakrilxji.chromatic.com/?path=/story/workspace-admin-practices-practice-setup--with-instance-catalog)
2. [The catalog list — each row carries its reason](https://66a8981a27ced8fef3190d41-orakrilxji.chromatic.com/?path=/story/workspace-admin-practice-adoption-available-practices--default)
3. [Reviewing a practice before adding it](https://66a8981a27ced8fef3190d41-orakrilxji.chromatic.com/?path=/story/workspace-admin-practice-adoption-practice-panel--available)
4. [Reading one of your own](https://66a8981a27ced8fef3190d41-orakrilxji.chromatic.com/?path=/story/workspace-admin-practices-workspace-practice--default)

**The reading order**
- [Rationale leads, the rule follows](https://66a8981a27ced8fef3190d41-orakrilxji.chromatic.com/?path=/story/workspace-admin-practice-adoption-definition-preview--rationale-leads-the-rule-follows) · [Criteria as markdown](https://66a8981a27ced8fef3190d41-orakrilxji.chromatic.com/?path=/story/workspace-admin-practice-adoption-definition-preview--criteria-is-markdown) · [Only one guidance field](https://66a8981a27ced8fef3190d41-orakrilxji.chromatic.com/?path=/story/workspace-admin-practice-adoption-definition-preview--rationale-only)

**Groups**
- [Adding a whole group](https://66a8981a27ced8fef3190d41-orakrilxji.chromatic.com/?path=/story/workspace-admin-practice-adoption-area-panel--mixed-outcomes) · [Restoring a removed one](https://66a8981a27ced8fef3190d41-orakrilxji.chromatic.com/?path=/story/workspace-admin-practice-adoption-area-panel--restore-deleted-area) · [Choosing its icon](https://66a8981a27ced8fef3190d41-orakrilxji.chromatic.com/?path=/story/workspace-admin-practices-area-details--choosing-an-icon)

**Shared pieces**
- [Work type](https://66a8981a27ced8fef3190d41-orakrilxji.chromatic.com/?path=/story/shared-practice-vocabulary-work-type--every-kind) · [Autonomy source](https://66a8981a27ced8fef3190d41-orakrilxji.chromatic.com/?path=/story/shared-practice-vocabulary-autonomy-source--chosen-here) · [Meta row](https://66a8981a27ced8fef3190d41-orakrilxji.chromatic.com/?path=/story/common-meta-row--default) · [Filter toggle](https://66a8981a27ced8fef3190d41-orakrilxji.chromatic.com/?path=/story/common-filter-toggle--default) · [Status badge gallery](https://66a8981a27ced8fef3190d41-orakrilxji.chromatic.com/?path=/story/shared-practice-vocabulary-status-badge--every-registry)

**Failure and edges**
- Stale plan, now identical for both: [group](https://66a8981a27ced8fef3190d41-orakrilxji.chromatic.com/?path=/story/workspace-admin-practice-adoption-area-panel--plan-went-stale) · [practice](https://66a8981a27ced8fef3190d41-orakrilxji.chromatic.com/?path=/story/workspace-admin-practice-adoption-practice-panel--preview-went-stale)
- [Catalog failed to load](https://66a8981a27ced8fef3190d41-orakrilxji.chromatic.com/?path=/story/workspace-admin-practices-practice-setup--instance-catalog-failed) · [Filtered to nothing](https://66a8981a27ced8fef3190d41-orakrilxji.chromatic.com/?path=/story/workspace-admin-practices-practice-setup--filtered-to-nothing) · [Name unavailable](https://66a8981a27ced8fef3190d41-orakrilxji.chromatic.com/?path=/story/workspace-admin-practice-adoption-practice-panel--name-unavailable)
- At 320px: [practice](https://66a8981a27ced8fef3190d41-orakrilxji.chromatic.com/?path=/story/workspace-admin-practice-adoption-practice-panel--narrow-viewport) · [group](https://66a8981a27ced8fef3190d41-orakrilxji.chromatic.com/?path=/story/workspace-admin-practice-adoption-area-panel--narrow-viewport) · [workspace practice](https://66a8981a27ced8fef3190d41-orakrilxji.chromatic.com/?path=/story/workspace-admin-practices-workspace-practice--narrow-viewport)

A running application is behind the **Preview / Coolify** check.


FelixTJDietrich:
## Maintainer sequencing review after #1442

The explicit browse/preview/adopt experience is strong: ETag concurrency, duplicate protection, auditing, provenance, and the “Not yet validated” state should stay. CI is green.

Please **hold this PR for coordinated rework** before merge:

1. **Use #1356's single rollout state.** This branch adds practice.delivery_mode beside practice.is_active, creating combinations that the decided OFF | SHADOW | ACTIVE state intentionally removes. #1356 should own that canonical state; adoption starts at SHADOW.
2. **Do not put delivery policy on Observation.** observation.delivery_mode plus pervasive delivery_mode = ACTIVE query filters hides shadow evidence at the semantic/read layer. Shadow/non-delivery belongs in the resolver/feedback decision trace so #1358 can preview and explain it.
3. **Fix the historical occurrence-key eligibility leak.** PracticeDetectionDeliveryService calls findActiveOccurrenceKeys(workspaceId, occurrenceKeys) across historical observations. After a future Active → Shadow transition, an older ACTIVE row with the same occurrence key can make the current shadow result eligible (the insert conflict preserves that historical row). Resolve eligibility for the current event/ledger decision, not by historical occurrence key.
4. **Rebase after #1444 and #1356.** #1444 overlaps the practice fingerprint/schema, and #1356 owns delivery semantics. Preserve #1442's suppression/egress behavior while resolving the shared review handlers.

This is not a request to reduce the adoption UX; it is to land it once, on the durable rollout/evidence model rather than immediately migrating away from an interim one.

PR #1521, opened 2026-08-24
fix(server): let the review write its own opening

FelixTJDietrich:
## Description

A review comment opened with one of a small set of server-authored lines — `Nice work …` for a clean change, `Worth keeping: …` for a strength beside a problem — picked out of a curated slug-to-phrase map. Every review that fell in the same bucket therefore opened with the same sentence, and a change with a critical finding in it could still open on praise. The opening now comes from the review itself: the composer calls a new `report_summary` tool with one or two sentences about the change as a whole, the runner carries them out in the feedback envelope as `lead`, and the note opens on them.

Nothing stands in for a missing one. With no lead the note opens on its first finding, which tells the reader more than a stock pleasantry does.

### What moved, and what did not

Three constraints lived inside the phrase maps because a fixed sentence could satisfy them by construction. They now live in `feedback-composer.md`, which is where the judgement they encoded actually happens:

- an opening never implies a verdict on merging (no *before merging*, no *blocking*);
- it never leads with praise in front of a serious problem;
- it never counts or re-lists what follows.

Worth being explicit that this is a real trade: those were code-level guarantees and they are now prompt-level ones. The bound the server keeps is the one it cannot delegate — the review's text is re-clamped to the length budget here rather than trusted from the runner, and scrubbed of grading vocabulary, so a lead that empties out opens nothing rather than a blank line.

The count header went with the opener: the list underneath already carries the count, and the one number the reader cannot recover from what is shown — how many minor suggestions were collapsed — is still stated.

### Coherence bound

`requireMatchingCompositionDigest` guarantees the composer saw the admitted observation set, so the lead is written over the observations that are delivered. Two later filters (the in-context severity gate and reaction suppression) can still narrow that set after the lead is written, so a lead that named a specific finding could in principle outlive it. The prompt steers the lead at the change rather than at the list, which keeps that bounded; it is not eliminated.

## How to test

`DeliveryComposerTest` and `FeedbackCompositionResultParserTest` cover the seam:

- the lead is purely additive — the note with a lead is exactly the lead plus the note without one, which pins both the rendering and the absence of any substitute opener;
- a lead that scrubs to nothing opens nothing, not a blank line;
- blank, whitespace-only, absent and `null` leads all parse to no lead;
- every blocker survives the cap and the collapsed minor tail still says what it is hiding.

Full suite locally: `pnpm run check` (exit 0), 6720 server tests, 67 agent specs, all green.

## Checklist

- [x] If the operator must act on this change (new required env var, manual migration step), the changeset summary says how (`**Operators:** …`) and `MIGRATION.md` is updated — no operator action; `patch`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01LK2yxSEfNL5Q5xwbAqAMWr


FelixTJDietrich:
Folded into #1519 — the two are one change now, reviewed and cut down together.

PR #1575, opened 2026-08-29
docs(ci): say that preview sign-in serves one preview at a time

FelixTJDietrich:
## Description

Sign-in on a preview is confirmed working end to end — real staging data behind a real GitHub login. But the callback URL contains the PR number (`pr1538.api.<zone>/login/oauth2/code/github`), and GitHub rejects wildcard callbacks, so **one preview at a time** can be signed into until an operator re-points it.

`docker/preview/.env.example` already says this to operators. `ci-cd.mdx` — the page a contributor reads before labelling their PR — said only that sign-in works when the OAuth app is configured, which reads as a promise. The first contributor to try it on their own preview would get GitHub's `redirect_uri` error with no explanation where they were looking.

Worth landing before previews are announced to the team.

## How to test

CI covers this — documentation only. The behaviour it describes was verified by signing in to the live preview for #1538.

## Checklist

- [x] My changeset summary reads as an operator/user-facing note — empty changeset: documentation only
- [x] If the operator must act on this change, the changeset says how — no operator action

### DIFF
--- a/.changeset/preview-signin-caveat.md
+++ b/.changeset/preview-signin-caveat.md
@@ -1,0 +1,4 @@
+---
+---
+
+Contributor documentation only; no deployed behaviour changes.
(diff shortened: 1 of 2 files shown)


FelixTJDietrich:
Withdrawn — built on a claim that is wrong. Wildcard callbacks do work for this app, so preview sign-in is not limited to one preview at a time and contributors need no operator involvement. Replacing this with a fix to the inherited claim in `docker/preview/.env.example` that said the opposite.

